PR #17, opened 2018-09-18
Feature/slt 134 postupgrade split

Jancis:
**Story:**
https://wunder.atlassian.net/browse/SLT-134

**Description:**
splits post-install resource into two - one that gets triggered after new helm deployment install and another that is run after deployment update.

- Install checks for config files in sync folder. If they exist, installs site with a `config_installer` profile and imports said configuration. ~imports database dump from (shared) mounted location `/var/backups/db`. Shared NFS location must be deployed beforehand.~
 - Update imports config files, updates entitity configs and runs database updates.

---

Shared persistent volume resource for project
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: {{ .Release.Namespace }}-dbdump
  namespace: {{ .Release.Namespace }}
spec:
  storageClassName: nfs
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 10Gi
```

### DIFF
--- a/chart/templates/postinstall.yaml
+++ b/chart/templates/postinstall.yaml
@@ -20,7 +20,7 @@ spec:
         {{ include "drupal.php-container" . | indent 8 }}
         command: ["/bin/sh", "-c"]
         args:
-        - {{ .Values.postinstall.commands | quote }}
+        - {{ .Values.drupal.postinstall.command | quote }}
       imagePullSecrets:
       - name: gcr
       volumes:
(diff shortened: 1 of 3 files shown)


Jancis:
- Made a split, removed backup restore on first deployment, just the site-install with config install if such exists. 
- Moved configs to values file, based on suggestions from https://github.com/wunderio/charts/pull/3/files

PR #29, opened 2018-10-08
SLT-99 configurable cron

floretan:
Based on the work done by Agnis in https://github.com/wunderio/drupal-project-k8s/tree/feature/SLT-99-cron and by Jonathan in https://github.com/wunderio/charts/pull/3/files

We want to enable multiple cron jobs, for example, one for `drush cron`, and one for a dedicated queue. The use of the root scope `$` greatly simplifies the use of global variables in a loop.


Jancis:
Tested by adding several cron entries to `values.yaml` and running `helm install --dry-run --debug  ./chart/`. Works correctly, generates requiested cron resources.

PR #35, opened 2018-10-16
SLT-138: Make resources configurable.

floretan:
Previously, only the nginx server had resources set, otherwise the defaults come into play (10% of a  CPU unit requested for each container). We want to make the resource allocation configurable, but also to set lower default to be able to have more environments on the same cluster.


floretan:
Looking at the ticket, I realize I'm missing storage space and storage classes.

PR #36, opened 2018-10-16
Define app name as a helper

jonathanjfshaw:
Gitlab offers terminal access to K8s environments from the Gitlab UI, and UI feedback on deployments but only if  "app" has a specific value `app=$CI_ENVIRONMENT_SLUG`.

So how about allowing parent charts to customise the app name if they want to?

### DIFF
--- a/chart/templates/_helpers.tpl
+++ b/chart/templates/_helpers.tpl
@@ -4,10 +4,6 @@ version: {{ .Chart.Version }}
 release: {{ .Release.Name }}
 {{- end }}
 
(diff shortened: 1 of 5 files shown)


floretan:
Adding this metadata definitely doesn't hurt, I've seen other charts using the "app" field as well. Shouldn't we make this a value instead of a template?

jonathanjfshaw:
I'm not sure which approach is preferable; either will work for me.

The template allows for advanced logic like sanitisation as you do for the domain name; but it requires a parent chart to customise it.

PR #37, opened 2018-10-16
Stray reference to gcr in postupgrade

jonathanjfshaw:
This one got added in a parallel branch, it missed the cleanup of the gcr imagePullSecrets.


floretan:
Thank you. I completed the tests to cover that and will add them as a follow up.

PR #38, opened 2018-10-18
Make mariadb env variables optional

jonathanjfshaw:
Currently the database values are always assumed to be coming from the mariadb values. 

But we can allow for the possibility that these are set directly in values.drupal.env. We simply need to NOT automatically write the database variables from mariadb if mariadb is disabled.

Without this setting env variables like "DB_USER" via values.drupal.env will cause duplicate environment variables, which can be problematic.

### DIFF
--- a/chart/templates/_helpers.tpl
+++ b/chart/templates/_helpers.tpl
@@ -64,6 +64,7 @@ imagePullSecrets:
 {{- end }}
 
 {{- define "drupal.env" }}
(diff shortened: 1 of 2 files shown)


jonathanjfshaw:
I couldn't get your version to work, but I wrote a test so you can play with it more easily.

floretan:
Thank you @jonathanjfshaw!

The test was very helpful, I adjusted your code and committed it (see https://github.com/wunderio/drupal-project-k8s/pull/39)

PR #43, opened 2018-10-24
Leave random password generation to the mariadb chart

jonathanjfshaw:
If I don't explicitly specify a password in my parent values.yaml, my drupal installation fails.
Running printenv inside the dysfunctional drupal pod I see:
`DB_PASS={{ randAlphaNum 10 }}`

Which seems wrong, and to come from your values.yaml. I think the problem is that values.yaml is not processed as template, so `{{ randAlphaNum 10 }}` is being treated as a literal password, and something down the line can't handle a password with spaces in it.

It's also unnecessary to have this in values, we can leave it to the mariadb chart to handle random generation.

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -62,11 +62,13 @@ mariadb:
   replication:
     enabled: false
   rootUser:
-    password: "{{ randAlphaNum 10 }}" # pass in a value to your helm chart to override this
+    # A random password will be generated if you do not supply one. 
+    password:
   db:
     name: drupal
     user: drupal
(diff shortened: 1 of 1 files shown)


floretan:
I think these fields can even be completely from the values file. I'm even tempted to set the `rootUser.forcePassword` to true, because the random password option breaks as soon as a helm release is updated (another random password is generated, and the database can't boot).

floretan:
Thank you, the change has been integrated as part of our restructuring of the values.yaml file.

PR #44, opened 2018-10-25
Update Janis' release deletion script to support namespaces.

floretan:


jonathanjfshaw:
I'm finding that these nfs pv's get stuck not uncommonly in an undeleteable state (no idea why). Even `kubectl delete --now` or 'kubectl delete --force' don't help. To fix I use:
```
    - PV=$(kubectl get pvc ${CI_COMMIT_REF_SLUG}-public-files --namespace ${KUBE_NAMESPACE} --output=json | jq -j '.spec.volumeName') || true
    - kubectl patch pv $PV -p '{"metadata":{"finalizers":null}}' --namespace ${KUBE_NAMESPACE} || true
    - kubectl patch pvc ${CI_COMMIT_REF_SLUG}-public-files -p '{"metadata":{"finalizers":null}}' --namespace ${KUBE_NAMESPACE} || true
    - kubectl patch pod --selector=app=${CI_ENVIRONMENT_SLUG} -p '{"metadata":{"finalizers":null}}' --namespace ${KUBE_NAMESPACE} || true
```

floretan:
@jonathanjfshaw we also had a few of these undeletable PV's, and it turned out that there were old jobs (leftovers from cron jobs or post-install hooks) that were not active anymore, but they "used" the volumes anyway. Getting rid of them made it possible to delete the volumes, no `--force` or `--grace-period=0` was needed.

PR #49, opened 2018-10-30
SLT-146: Add a README file to the chart explaining how it is used.

floretan:
Rather than documenting all the options in a markdown table, I think it's easier to point people directly to the values.yaml file where there are now some comments for all the options.

### DIFF
--- a/chart/README.md
+++ b/chart/README.md
@@ -47,9 +47,22 @@ $ helm upgrade --install drupal
     
 ```
 
-All default values are are optimised for low resource usage on non-production environments. Production environments 
-should have a dedicated files with values adapted to the project, among others: 
+All default values are are optimised for low resource usage on non-production environments.
+Production use cases will be supported (they have not been tested yet), but they should
+have a dedicated values file adapted to the needs of the project, among others:
 - High availability deployment using replicas.
 - More dedicated CPU and memory.
-- Database delegated to an external hosted service.
-- Dedicated ingress resource with a dedicated domain.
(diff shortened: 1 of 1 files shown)


jonathanjfshaw:
Also, there's no discussion here about what the cluster requirements are: ambassador, readWriteMany storage, etc. This is one of the topics that could be most helpful to document.

Do you have a public repo that has a chart for setting up a cluster? That's what I've started to do, partly for replicability, partly just to document what I've done for my future self.

floretan:
> Nit: your bullet point numbering looks broken.
Markdown turns any numbers into sequential bullets, but I guess it's nicer if the source file makes sense too.

> The wording is a bit confusing, as AFAIK you can't setup replicas or ingress from a dedicated values file; this needs a parent chart wrapping your chart.
I clarified the wording. The replicas for the main Deployment object can already be set in values.yaml, we might add autoscaling support later. A configurable ingress will be added to the chart, but we haven't needed it yet as our current goal is to get development environments running smoothly.

> HTTPS probably belongs on this list too.
Yes, this is on our list even for development environments and will most likely be addressed in the next sprints (starting in two weeks).

> Are you planning a public production chart to collaborate on this kind of thing?
The same chart should support production hosting as well. Once we start testing this use case, we will also provide a sample values file, but the exact configuration will change based on the requirements of each project.

jonathanjfshaw:
That all sounds great.

https://github.com/wunderio/silta-cluster gives me a 404, I think maybe you haven't made this public?

PR #50, opened 2018-10-31
The dockerfiles themselves should not be in the docker images.

floretan:


floretan:
http://feature-hide-dockerfiles.drupal-project-k8s.silta.wdr.io/Dockerfile is not present anymore.

PR #57, opened 2018-11-08
Feature/slt 154 ssh jumphost

Jancis:
**Story:** 
https://wunder.atlassian.net/browse/SLT-154

**Note: This PR should be accepted only when https://github.com/wunderio/sshd-gitauth and https://github.com/wunderio/silta-cluster/pull/2 is reviewed**
 
**What this PR does:**
- Adds (build and deploy) an ssh image to the project, based on the php image. It's because we need need php and executables to run drush commands. I did not even notice any build time penalty, it's still around 3 minutes.
- Adds a `branchnameUrl` variable to helm deployment. It's a dns-friendly of branch name. We use this for shell service, because we want nicer dns for the command we give to users.
- Another `RELEASE_NAME` rewrite, it's a `projectname-branchname` now and when either of those exceed 25 symbols, we truncate them and add a 4 chars long hash of the name. The chances of collision should be really low. We might, however, want to change all of resource names from `{{ .Release.Name }}` to `{{ .Values.branchnameUrl }}`, just have to look out for mariadb resource name (correct reference name and that the final helm release name is short enough for it)
- A set of variables in `values.yaml` and a python script that connects to github and reads public keys of the project members. This script is used as `AuthorizedKeysCommand` for the OpenSSH. The resulting member list is cached for an hour and the member keys are cached for 24 hours

There's also a related work done for this:
 - SSH docker image with github authentication: https://hub.docker.com/r/wunderio/sshd-gitauth and https://github.com/wunderio/sshd-gitauth
 - [SSH Jumphost for Silta cluster](https://github.com/wunderio/silta-cluster/pull/2)

**Testing**
Due to silta-cluster deployment issues, i've deployed only the jumphost resource. You can pass trough it, if you are member or external collaborator of `wunderio` organisation. Then the per-branch deployment resource, "`shell`", will allow you, if you are project collaborator.
So, try this command: `ssh www-admin@feature-slt-154-ssh-jumphost-shell.drupal-project-k8s -J root@35.197.223.124`

**Issues and Improvements:**
 - Assign dedicated hostname for the jumphost when the , `NOTES.txt` file suggests `ssh.silta.wdr.io`
 - All resource names could be set to branchname
 - Drupal environment variables for shell container. Drupal environment variables are not passed so the database connection (and possible something else) is unavailable. They are available with `kubectl exec`, but not accessible with ssh. I currently circumvent this by piping env variables to `~www-admin/.ssh/environment`, but there should be better solution as it also exposes information from other branches too. It's too noisy.
 - ~Random connection issues. Sometimes the connection would not initialize. Try 3 times and it succeeds. Then it works 10 times out of 10. This was way worse with `centos-7` as base image, but is better with alpine. Maybe lack of resources?~ 
```
channel 0: open failed: connect failed: Connection refused
stdio forwarding failed
ssh_exchange_identification: Connection closed by remote host
```
UPDATE: Found out the issue, the service was pointing at app drupal and it randomly pointed port 22 requests to wrong pods (web pods). They are separated now. 
 - When the container is recreated, ssh fingerprint does not match anymore
 - Limit gitauth login to www-admin. Currently the `AuthorizedKeysCommand` is set for all users, It does not matter much, but we should probably add a `Match User www-admin` section with `AuthorizedKeysCommand` instead.
 - Inject prefefined ssh keys as backup. Maybe.
 - Jumphost `drush/sites/self.site.yml` entry for drush commands, like this
```
feature-slt-154-ssh-jumphost:
  host: feature-slt-154-ssh-jumphost-shell.drupal-project-k8s
  user: www-admin
  root: '/var/www/html'
  uri: 'http://feature-slt-154-ssh-jumphost.drupal-project-k8s.silta.wdr.io'
  ssh:
    options: '-J root@ssh.silta.wdr.io'
```
  So it would be very easy to do stuff like `drush @feature-slt-154-ssh-jumphost uli` or even `drush rsync @master:%files @feature-slt-154-ssh-jumphost:%files`


silta/shell/Dockerfile
```diff
@@ -0,0 +1,17 @@
+# Dockerfile for the Drupal container.
+FROM wunderio/drupal-php-fpm:v0.1
+
+COPY --chown=www-data:www-data . /var/www/html
+USER www-data
+RUN mkdir -p -m +w /var/www/html/web/sites/default/files
```
floretan: This line is not needed anymore, the mount point is created in the base image.

Jancis:
@jonathanjfshaw Florian asked about the same part yesterday. I will just reuse the `environmentName` and trim it down if necessary.

Jancis:
Made a few changes: 
 - SSHD logs are now directed to `/proc/self/fd/2` which allows using `kubectl log`
 - `ssh_host_dsa_key` key added to remove error where sshd looks for the file on each connection
 - Direct services to correct deployments (22 to shell deployment and 80 to drupal), otherwise it tries to connect to the wrong pods. Now the SSH connection works 10 out of 10 times :tada: 
 - Shell `livenessProbe` removed to avoid getting constant `Did not receive identification string from XXX`.

Jancis:
Closed due to having some errors, but that turned out to be issue in master.

PR #58, opened 2018-11-08
SLT-166: Import reference database dump for new environments

floretan:
You can delete the `w68f3d122-b5a8b6f7-feature-reference-da` release and re-run the last deployment from https://circleci.com/gh/wunderio/drupal-project-k8s/tree/feature%2Freference-data-import, you will then get a working environment running on http://feature-reference-data-import.drupal-project-k8s.silta.wdr.io/ based on the latest reference database dump. Note that name of the admin user is set to a random generated value, thanks to the GDPR sanitization.

What is not included and moved to a dedicated ticket:
- Optimise the dumps so the cache tables are not included
- Copy the drupal ~public and~ private files 
- Skip dumps if the gdpr-dump package is not present

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,7 +98,6 @@ jobs:
             --set mariadb.db.password=$DB_USER_PASS \
             --namespace=${CIRCLE_PROJECT_REPONAME,,} \
             --values silta.yml \
-            --timeout=600 \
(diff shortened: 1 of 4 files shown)


Jancis:
Looks good, thank you for updating the creation part too!

PR #61, opened 2018-11-09
SLT-163: display logs from helm hooks

floretan:
That was surprisingly easy. Following the log from a job automatically stops when the log is complete. The consolidation of the post-install and post-upgrade into a single post-release job that conditionally uses the post-install or post-upgrade command also provides a nice reduction of duplicate code.

One thing to be addressed is that if the hook fails, we see the failure in the logs, but the CircleCI build itself would pass, so we might need an extra check for that.


Jancis:
Great! Extra hugs for reduction of code and usage of IsInstall object!

PR #65, opened 2018-11-14
Feature/slt 183 postinstall db readiness check

Jancis:
**Story:** 
https://wunder.atlassian.net/browse/SLT-183

**Description:**
> When a helm release is created without the `--wait` parameter, the post-install hook gets executed right away, before the database is ready. The post-install hook should poll the database using `drush status --field=Database` and wait for the output to be `Connected`.

Tried to use both `drush status --field=Database` and the new `drush check-bootstrap:db`, but it returned `[notice] Missing database table: key_value` so i went with a raw database connection.

I'm open to any suggestion of making this a bit less ugly.


Jancis:
`--connect_timeout=2` is there just to limit connection time when the database is still unreachable.
 
>  The maximum number of seconds before connection timeout. The default value is 43200 (12 hours). 
https://dev.mysql.com/doc/refman/8.0/en/mysqladmin.html#option_mysqladmin_connect_timeout

PR #70, opened 2018-11-20
Feature/slt 186 make ssh optional

Jancis:
- Enables shell container (SSH) only when `values.yaml` `shell.enabled` is set to `true`. It's off by default now   
- Some tests, that check the `enabled` setting and that required values are correctly passed as environment variables into shell container.  
- Even when the shell **is** enabled it is not set when the `referenceEnvironment` equals `environmentName` (production, master). This also means, first deployment to master can't be fixed with a db push trough shell container, another branch has to be created for it. So maybe the initial setup should be changed somehow. This is only because some projects might still want to have shell, but there is no way to do it per-branch.

### DIFF
--- a/chart/templates/NOTES.txt
+++ b/chart/templates/NOTES.txt
@@ -7,6 +7,7 @@ Basicauth username: {{ .Values.nginx.basicauth.credentials.username }}
 Basicauth password: {{ .Values.nginx.basicauth.credentials.password }}
 {{- end }}
 
+{{- if and (.Values.shell.enabled) (ne .Values.referenceData.referenceEnvironment .Values.environmentName) }}
 SSH connection:
 
   ssh www-admin@{{ template "drupal.environment.hostname" . }}-shell.{{ .Release.Namespace }} -J www-admin@ssh.{{ .Values.clusterDomain }}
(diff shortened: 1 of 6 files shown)


Jancis:
Maybe shell resource can be built and deployed when the commit is tagged with specific tag, i.e `enable_shell`

chart/templates/NOTES.txt
```diff
@@ -7,10 +7,12 @@ Basicauth username: {{ .Values.nginx.basicauth.credentials.username }}
 Basicauth password: {{ .Values.nginx.basicauth.credentials.password }}
 {{- end }}
 
+{{- if and (.Values.shell.enabled) (ne .Values.referenceData.referenceEnvironment .Values.environmentName) }}
```
floretan: I would only check for `.Values.shell.enabled`, otherwise it's impossible to get a shell container for the reference environment. If the reference environment is the production environment, we would have a dedicated configuration file, and we could set `shell.enabled: false` there.

chart/values.yaml
```diff
@@ -132,6 +132,7 @@ php:
 
 # Configuration for everything that runs in shell container.
 shell:
+  enabled: false
```
floretan: Most of the settings are optimised for development environments, and in that context a shell container makes a lot of sense, so I would have it enabled by default.

One of the reasons for not having a shell container enabled by default would be to speed up the deployment. In that case we should also prevent the container from being built in the first place, but that's definitely a different task.

chart/templates/NOTES.txt
Jancis: Fixed

chart/values.yaml
Jancis: reenabled

PR #74, opened 2018-11-23
SLT-91: Elasticsearch

agnis-mateuss:
**Story:** https://wunder.atlassian.net/browse/SLT-91

**What this PR does**
Adds Elasticsearch template, which does the following:

- Installs ES 6.5 as a single container pod.
- Exposes Elasticsearch service as `ELASTIC_HOST` environment variable
- Implements workarounds to allow ES execute with usable performance. ( See [elasticsearch.yml](https://github.com/wunderio/drupal-project-k8s/blob/60fb59c24a63d1aebaedbe520afbc17cb50dffdc/chart/templates/elasticsearch.yaml#L32) )

**Reasoning for custom template**
Chart from `helm/stable/elasticsearch` only provides setup for clusters, and not single node installations. At ~700MB size image, it's also costly for our use case.  
Second issue - full resource names, coupled with Helm release name, were longer than 64 characters.

**Testing**

1. Deploy release with elasticsearch.enabled: true
2. curl the elasticsearch service with `curl ${ELASTIC_HOST}:9200` from shell or drupal-php container
3. For persistence:
    1. post data to elastic (`curl -H "Content-Type: application/json" -XPOST ${ELASTIC_HOST}:9200/test/test -d '{"message":"test","sample":"lorem ipsum"}'`
    2. delete elasticsearch pod
    3. query elastic service (`curl -X GET "${ELASTIC_HOST}:9200/_all/_search?q=message:test"`). Slight waiting time before new pod gets created

**Issues and Improvements**

- Try out service with multiple replicas
- Need to keep an eye on performance for anomalies


chart/templates/elasticsearch.yaml
```diff
@@ -0,0 +1,98 @@
...
+      #https://github.com/kubernetes/kubernetes/issues/3595#issuecomment-423253874
+      securityContext:
+        fsGroup: 1000
+        capabilities:
+          add:
+            - IPC_LOCK
+            - SYS_RESOURCE
+      containers:
+      - name: elastic
+        image: docker.elastic.co/elasticsearch/elasticsearch:6.5.0
```
Jancis: Would be great to be able to define version in `values.yaml`. Add a link (https://www.docker.elastic.co/) to available images as a comment too.

agnis-mateuss: Implemented in https://github.com/wunderio/drupal-project-k8s/pull/74/commits/0afa35000af306687a4b3620d300f06cb7502fe8 .

chart/templates/_helpers.tpl
```diff
@@ -8,6 +8,11 @@ app: shell
 release: {{ .Release.Name }}
 {{- end }}
 
+{{- define "elastic.release_labels" }}
```
floretan: The prefix in these template definitions should match the chart name, so in this case it should be `drupal`, not `elastic` (by the way, Elastic is the company, Elasticsearch is the software).

chart/templates/_helpers.tpl
```diff
@@ -8,6 +8,11 @@ app: shell
...
+app: elastic
```
floretan: I would rather use the existing `drupal.release_labels` and extend those in the deployment definition, the same way that we are doing it for the Drupal deployment.

chart/templates/_helpers.tpl
agnis-mateuss: https://github.com/wunderio/drupal-project-k8s/pull/74/commits/041607b1c4bfdd62a4aba1f4f949d6c4fdf7130c  
Inserted common label set (app,release) for shell, elasticsearch resources. For selectors to differentiate between these, I added third label `service` to mark which accompanying service it is, i.e. `service: shell`

chart/templates/_helpers.tpl
agnis-mateuss: Same as https://github.com/wunderio/drupal-project-k8s/commit/041607b1c4bfdd62a4aba1f4f949d6c4fdf7130c

chart/values.yaml
```diff
@@ -196,6 +196,18 @@ mariadb:
     persistence:
       size: 1G
 
+elasticsearch:
+  enabled: true
```
floretan: Now that we know this works, we should keep elasticsearch disabled by default.

chart/values.yaml
```diff
@@ -196,6 +196,18 @@ mariadb:
...
+  #Choose versions in https://www.docker.elastic.co/ from "Elasticsearch" list
+  version: 6.5.0
+  persistence:
+    size: 512Mi
+    storageClassName: nfs
+  requests:
```
floretan: We should pass in a whole `resources` parameter, so that we can add not only requests but also limits, this will be needed when we tune things, especially given how Elasticsearch tends to reserve any resources it can get its hands on.

chart/values.yaml
```diff
@@ -196,6 +196,18 @@ mariadb:
...
+    cpu: 25m
+    memory: 512Mi
+    heapSize: 256m
```
floretan: The heapSize is a parameter for the resources _within_ the container, it should not be mixed with resource configuration for the container itself.

PR #76, opened 2018-11-28
SLT-198: Drupal service should only connect to pods from deployment.

floretan:


floretan:
After a deployment, http://bugfix-slt-198-service-label-selector.drupal-project-k8s.silta.wdr.io returned the full page right away, no more upstream errors.

PR #79, opened 2018-12-03
SLT-201: Update reference data after deployments of the reference environment.

floretan:


floretan:
Note: this cannot be tested as the feature branch is not the reference environment.

chart/values.yaml
```diff
@@ -159,7 +159,13 @@ referenceData:
   # The name of the environment from which reference data will be copied.
   referenceEnvironment: 'master'
 
+  # When to automatically update the reference data.
   schedule: '0 3 * * *'
+
+  # Update the reference data after deployments on the reference environment.
+  afterDeployment: true
```
floretan: Maybe `updateAfterDeployment` is a clearer term.

PR #81, opened 2018-12-06
SLT-187: Mount ssh keys onto persistent storage that lives trough deployments

Jancis:
Moves ssh fingerprint keys into separate mounted folder


Jancis:
all new deployments are broken because this is missing. merging in on my own

PR #82, opened 2018-12-10
Readiness improvements

floretan:
Based on the load tests, we see two necessary improvements for the readiness checks:

- The readiness checks should be done more frequently (the default is every 10 seconds), so once Drupal is installed it takes less time for the pod to be marked as ready and for the service to be active.
- Under load, Drupal can take longer to respond, exceeding the default liveness and readiness timeout of 1 second. While this is not ideal, such conditions can commonly occur after a cache rebuild, so we need to be more tolerant.


floretan:
These changes need some additional validation, and will be merged together with the load tests.

PR #88, opened 2019-01-16
Feature/slt 220 fix backoff limit exceeded

Jancis:
- Creates temporary file while site-install is running and prevents php requests and cron commands from being run. It's because I suspect some drush actions create tables in database while site-install is running, hence breaking the deployment. And because backofflimit is set to 0, it fails right away.
- Waits for reference FS to be ready by trying to touch a file on it. It's only set up to work on `.Values.referenceData.referenceEnvironment`. Maybe it should work every time to be more reliable, but then the storage can't be read-only for non-`referenceEnvironment` environments.

### DIFF
--- a/chart/templates/_helpers.tpl
+++ b/chart/templates/_helpers.tpl
@@ -156,6 +156,10 @@ done
 rm /var/www/html/reference-data/_fs-test
 {{- end }}
 
+{{- define "drupal.deployment-in-progress-test" -}}
+-f /var/www/html/web/sites/default/files/_deployment
+{{- end -}}
+
 {{- define "drupal.post-release-command" -}}
 set -e
 
@@ -166,7 +170,9 @@ set -e
 {{ include "drupal.wait-for-db-command" . }}
 
 {{ if .Release.IsInstall }}
+touch /var/www/html/web/sites/default/files/_deployment
 {{ .Values.php.postinstall.command}}
+rm /var/www/html/web/sites/default/files/_deployment
 {{ else }}
 {{ .Values.php.postupgrade.command}}
 {{ end }}
--- a/chart/templates/_helpers.tpl
+++ b/chart/templates/_helpers.tpl
@@ -142,8 +142,26 @@ TIME_WAITING=0
 done
 {{- end }}
 
+{{- define "drupal.wait-for-ref-fs-command" }}
+TIME_WAITING=0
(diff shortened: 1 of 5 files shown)


jonathanjfshaw:
I don't like the way the logic moves back and forwards between values.yml and template files with these commands. It means that when using values.yml to set/modifying the install command for a project, you need to understand the esoteric deployment file logic that's used in the templates.

Could we do this instead:
```
{{- define "drupal.post-release-command" -}}
set -e

{{ include "drupal.wait-for-db-command" . }}

{{ if .Release.IsInstall }}
touch /var/www/html/web/sites/default/files/_deployment
{{ .Values.php.postinstall.command}}
rm /var/www/html/web/sites/default/files/_deployment
{{ else }}
{{ .Values.php.postupgrade.command}}
{{ end }}
```

This way developers can use their normal drupal/devops knowledge to tinker with their projects install process in values.yml, without having to know all about this chart's particular internal logic.

Also, would it make sense to do something like:
{{- define "drupal.deployment-in-progress-test" }}
-f /var/www/html/web/sites/default/files/_deployment
{{- end }}
By centralising this test into a template helper it allows someone to override it.

Lastly, maybe it would also make sense to not run cron during an upgrade as well as an install? I could imagine a situation where running an upgrade hook and a cron script at the same time led to unforseen problems, leaving drupal data in a strange state.

Jancis:
1. I agree that the the whole values-helper-templates logic is quite tangled. I was looking into possible ways to block requests to `php` container other than setting that file but didn't find anything usable. Especially when the readinessProbe doesn't even like inverse logic for non-exec probes.

2. Having the `_deployment` touched in for the whole `.Release.IsInstall` switch would be okay too. It's only for postInstall which would only happen initially, once per branch?

3. I'd avoid `postupgrade` having this lock, because php container is blocked as well then. My general feeling is it should be avoided, something like `maintenance_mode` can be set instead. 

4. It's a good idea to extract `drupal.deployment-in-progress-test`. I can also implement `deploymentLock`, `deploymentUnlock` and `deploymentIsLocked` helpers (`deploymentIsLocked` being the `drupal.deployment-in-progress-test`).

Jancis:
@jonathanjfshaw changes implemented
@floretan can we merge this? There's still the storage change missing, it will be a different PR.

floretan:
It looks like the tests need to be updated.

PR #89, opened 2019-01-22
SLT-89: separate helm values for certain environments

floretan:
We now have an example value file for production environments. The resource allocation is slightly higher and we have minimal high availability (only two instances), but these are settings that can work for the master environment even during development. More complex changes that require additional work are commented out, like the use of an external database. Disabling of basic authentication is also commented out so that we don't accidentally make an environment publicly accessible. 

The configuration files to be used are then made configurable in the CircleCI orb definition.

Follow up to this task is: 

1. Update the orb in https://github.com/wunderio/silta-circleci to add the new parameter
2. Update https://github.com/wunderio/drupal-project with silta-prod.yml and the separate CircleCI workflow for the master branch.


Jancis:
can we move `silta.yml` and `silta-prod.yml` files in `silta/` folder while we're at it?

floretan:
Yes, that's a good idea.

PR #100, opened 2019-02-15
SLT-95 resource optimisation

floretan:
I did some testing, and if we have more resources for the post-release job (see https://github.com/wunderio/drupal-project-k8s/pull/99) we can reduce the requested memory for the other containers. 

As discussed in Slack, I would also recommend removing the readiness checks. They are creating a considerable load on the service by themselves and prevent us from being able to lower the resources without the checks failing due to timeouts. They have also been responsible for a few confusing issues for developers, in the end we are better off without them.

Note that the deployment is a bit slower because in this branch the post-release job doesn't yet get additional resources.

### DIFF
--- a/chart/templates/drupal-deployment.yaml
+++ b/chart/templates/drupal-deployment.yaml
@@ -33,13 +33,6 @@ spec:
         livenessProbe:
           tcpSocket:
             port: 9000
-        readinessProbe:
-          exec:
-            command:
-            - /bin/sh
-            - -c
-            - if [ ! {{ include "drupal.deployment-in-progress-test" . }} ]; then timeout -t 3 drush check-bootstrap; else exit 1; fi
-          periodSeconds: 3
         resources:
           {{- .Values.php.resources | toYaml | nindent 10 }}
 
@@ -74,12 +67,6 @@ spec:
         livenessProbe:
           tcpSocket:
             port: 80
-        readinessProbe:
-          httpGet:
-            path: /robots.txt
-            port: 80
-          periodSeconds: 3
-          timeoutSeconds: 3
         resources:
           {{- .Values.nginx.resources | toYaml | nindent 10 }}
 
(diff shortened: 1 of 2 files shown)


chart/templates/drupal-deployment.yaml
```diff
@@ -33,13 +33,6 @@ spec:
         livenessProbe:
           tcpSocket:
             port: 9000
-        readinessProbe:
-          exec:
-            command:
-            - /bin/sh
-            - -c
-            - if [ ! {{ include "drupal.deployment-in-progress-test" . }} ]; then timeout -t 3 drush check-bootstrap; else exit 1; fi
```
Jancis: I believe that test is still needed so we don't get the tables created while post-deployment doing initial `drush site-install`.

floretan: You are right. We can skip `drush check-bootstrap` but should still check for the deployment in progress.

chart/templates/drupal-deployment.yaml
```diff
@@ -35,11 +35,7 @@ spec:
             port: 9000
         readinessProbe:
           exec:
-            command:
-            - /bin/sh
-            - -c
-            - if [ ! {{ include "drupal.deployment-in-progress-test" . }} ]; then timeout -t 3 drush check-bootstrap; else exit 1; fi
-          periodSeconds: 3
+            command: ['/bin/sh', '-c', 'if [ {{ include "drupal.deployment-in-progress-test" . }} ]; then exit 1; fi']
```
ragnarkurmwunder: To be more verbose I suggest:
`/bin/sh -x -c ...`
Extra log saves time when things go wrong.

Jancis: Good suggestion @ragnarkurmwunder . I believe this issue might be in several places, could you take a look at the code and make a PR for it, please? I'll try to bring it over to https://github.com/wunderio/charts when we test if on k8s

ragnarkurmwunder: Actually, you can simplify it too :)
```
{{ include "drupal.deployment-in-progress-test" . }} && exit 1
```

PR #107, opened 2019-03-19
SLT-234 basic checks

floretan:
A set of simple checks regarding issues we have seen in the past (generally about certain key files missing from the repository).


Jancis:
should we check for the nginx robots.txt file too?

floretan:
The robots.txt file is not used in the readiness check anymore, so nothing would fail if it's not present. The file is also not part of the base repository since it gets added with the initial `composer install`, which would require a workaround.

PR #108, opened 2019-03-20
SLT-209: Do not share shell key and elasticsearch storage

Jancis:
Story: https://wunder.atlassian.net/browse/SLT-209
Changes elasticsearch and shell key storage to standard

Related PR in `charts`: https://github.com/wunderio/charts/pull/19

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -129,7 +129,7 @@ shell:
 
   # Persistent storage for SSH fingerprint keys
   storage: 1M
-  storageClassName: silta-shared
+  storageClassName: standard
 
   # The docker image to be used. This is typically passed by your CI system using the --set parameter.
   image: 'you need to pass in a value for shell.image to your helm chart'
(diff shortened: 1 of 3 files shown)


Jancis:
Have to test how existing deployments work with this definition change

chart/values.yaml
```diff
@@ -129,7 +129,7 @@ shell:
 
   # Persistent storage for SSH fingerprint keys
   storage: 1M
-  storageClassName: silta-shared
+  storageClassName: standard
```
floretan: Rather than hard-coding "standard" (which is the default for Google Cloud but might not be elsewhere), we should leave that undefined, meaning using the default. See https://github.com/helm/charts/blob/d05341dea09eca4a88177988355b1a2af7abace9/stable/minio/values.yaml#L74 for a good explanation.

Jancis: Variables and definition removed

PR #113, opened 2019-06-24
SLT-217: Use local phpcs 

floretan:
We have had some cases where an updated global version of phpcs caused issues. It would be safer to use a project-specific version of phpcs (which turns out to be already installed as a dev dependency of Drupal core). 

We don't want to remove the globally installed phpcs yet as that could break a lot of sites, but it's nice to use the local one if present already.

As a follow-up, these changes need to be integrated in the silta-circleci orb.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -198,6 +198,7 @@ orbs:
               command: |
                 if [ -f vendor/bin/phpcs ]
                 then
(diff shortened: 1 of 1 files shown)


floretan:
Closed in favor of https://github.com/wunderio/silta-circleci/pull/20

PR #115, opened 2019-07-04
SLT-203: Dynamic volume generation

Jancis:
Dynamic file mounts. File mount is defined by default, can override "mounts" section to add more.

```
mounts:
  - name: public-files
    storage: 1G
    mountPath: /var/www/html/web/sites/default/files
    storageClassName: silta-shared
    csiDriverName: csi-rclone
  - name: private-files
    storage: 1G
    mountPath: /var/www/html/private
    storageClassName: silta-shared
    csiDriverName: csi-rclone
    readOnly: false
  - name: custom-mount
    storage: 1G
    mountPath: /var/www/html/custom-mount
    storageClassName: silta-shared
    csiDriverName: csi-rclone
    readOnly: false
```
Tests changed accordingly.

No storage migration neccessary - i kept the naming schema `drupal-public-files` and `drupal-private-files` when they are generated. Also, the `PRIVATE_FILES_PATH` env variable also works. All mounts have their env variables now.

Things to consider:
 - Legacy definitions of `.Values.publicFiles` and `.Values.privateFiles`. Maybe we need a temp resource that prints a message if anything is defined in those. Or not. Public is on by default now anyway.
 - Defining `readOnly` could be unnecessary in that scope. nginx pod could just set all dynamic mounts to false by default.
 - Probably ommiting `csiDriverName`. It could be just injected into PV, assuming that we're only going to use `csi-rclone` driver for storage. Or defaulting to `csi-rclone`.
 - Nginx now mounts all storages. It mounted only the public storage before.

### DIFF
--- a/silta/silta.yml
+++ b/silta/silta.yml
@@ -3,27 +3,3 @@
 #
 # See https://github.com/wunderio/charts/blob/master/drupal/README.md
 # for all possible options.
-
-# Example: enable private files for Drupal.
-#privateFiles:
-#  enabled: true
-
-# Testing mounts
-mounts:
-  - name: public-files
-    storage: 1G
-    mountPath: /var/www/html/web/sites/default/files
-    storageClassName: silta-shared
-    csiDriverName: csi-rclone
-  - name: private-files
-    storage: 1G
-    mountPath: /var/www/html/private
-    storageClassName: silta-shared
-    csiDriverName: csi-rclone
-    readOnly: false
-  - name: custom-mount
-    storage: 1G
-    mountPath: /var/www/html/custom-mount
-    storageClassName: silta-shared
-    csiDriverName: csi-rclone
-    readOnly: false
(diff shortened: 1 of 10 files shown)


silta/silta.yml
```diff
@@ -7,3 +7,23 @@
 # Example: enable private files for Drupal.
 #privateFiles:
 #  enabled: true
+
```
floretan: It's good to have tested this, but it should be removed before merging.

silta/silta.yml
```diff
@@ -7,3 +7,23 @@
 # Example: enable private files for Drupal.
 #privateFiles:
```
floretan: This example needs to be removed, or at least updated. There's an example in values.yaml, so we might not need anything here after all. This also needs to be updated in the drupal-project repo.

Jancis:
- `drupal` chart PR: wunderio/drupal-project-k8s#115
- `drupal-project` doc change: wunderio/drupal-project@85a15b1

PR #116, opened 2019-07-05
SLT-268: Change the application root to /app.

floretan:
The base docker images have been changed so that `/app` is a symlink to `/var/www/html`. Renaming the paths in the chart should work without issues. When all projects have been updated, the symlink will be removed.


Jancis:
Everything seems to be covered :+1:  
We'll need to remove `drupal.wait-for-ref-fs-command` at some point, i think it's obsolete.

PR #123, opened 2019-07-17
Refactor GDPR-dump

floretan:
This PR moves the GDPR-dump configuration from a dedicated file to silta.yml, which is then mounted as a mysql configuration file, which works independently of the drush version (drush 8 does not support the `--extra-dump` parameter). 

Benefits:
- less configuration files
- possibility to centrally update default rules
- should work with Drupal 7


floretan:
Closing in favour of https://github.com/wunderio/drupal-project-k8s/pull/127

PR #124, opened 2019-07-17
SLT-285: Use objects instead of arrays in silta.yml for easier overrides

floretan:
We use arrays in few places in values.yaml file (cron tasks, whitelisted ip addresses, volume mounts). Helm can iterate over objects, so turning those arrays into objects with arbitrary keys still works the same way. The advantage is that it is then a lot easier to override individual parameters or to add additional items without needing to duplicate the entire array.


floretan:
@Jancis this PR probably makes more sense when looking at the commits individually. The first one is the main change, which is quite minimal. The second one is to take advantage of the new possibility to optimise the way mounts are defined in the tests.

PR #132, opened 2019-07-24
SLT-301: Enable autoscaling for the Drupal deployment

floretan:
Steps to test:

The deployment should only have one replica:
```
kubectl get deployment -n drupal-project-k8s drupal-project-k8s--feature-slt-301-drupal
```

Generate some load on the feature branch environment: 
```
siege -c10 --header="Authorization:Basic c2lsdGE6ZGVtbw==" http://feature-slt-301.drupal-project-k8s.silta.wdr.io/
```

Then, while siege is still running check with the first command again to see that the deployment scaled up to 3 replica.


Jancis:
Tested, works! Took less than 10 minutes to settle down after load testing.

PR #133, opened 2019-07-29
SLT-307: Add fastcgi configuration from WunderMachina.

floretan:
We ran into an issue where the `fastcgi_buffers` defaults were too low, something that commonly happens with D8 and for which we have an existing configuration in WunderMachina. There are more settings that could be very useful (for example hiding the `x-generator` header), it probably makes sense to have all of them.

### DIFF
--- a/chart/templates/drupal-configmap-nginx.yaml
+++ b/chart/templates/drupal-configmap-nginx.yaml
@@ -72,6 +72,58 @@ data:
     }     
 
   drupal_conf: |
+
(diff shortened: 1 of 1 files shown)


floretan:
After thinking about it some more, I'm wondering if this configuration shouldn't be added to the base docker image directly, since it doesn't actually change per project.

floretan:
Merging as this fixes issues with existing projects. I created a dedicated ticket to consolidate the configuration in the base docker image.

PR #135, opened 2019-07-29
SLT-305: Trust all subdomains of the cluster domain.

floretan:
Here's a minimal change to get sites working with existing entries for the Drupal trusted hosts.

A few more options to consider:
- We could use the full domain rather than just the cluster domain
- What do we do about additional exposed domains?
- Could we create settings.silta.php as a configMap, so it doesn't have to be present in each project? This could be very flexible, but also potentially confusing.


Jancis:
I'd say the `trusted_host_patterns` to wildcard because only predefined domains are routed trough the ambassador to correct pods.

web/sites/default/settings.silta.php
```diff
@@ -41,3 +41,8 @@
  * Generated twig files should not be on shared storage.
  */
 $settings['php_storage']['twig']['directory'] = '../generated-php';
+
+/**
+ * Make sure the dynamic environments are not blocked out as untrusted.
+ */
+$settings['trusted_host_patterns'][] = '^.*\.' . preg_quote(getenv('SILTA_CLUSTER_DOMAIN')) . '$';
```
agnis-mateuss: Should add newline

floretan:
I'll update this branch based on our discussion.

PR #141, opened 2019-08-07
Give a hint that VPN is needed for ssh access.

floretan:


Jancis:
This is quite specific to our environment

PR #159, opened 2019-09-19
SLT-237: set pod affinity

floretan:
- If possible, keep replicas of the same Drupal deployment on different nodes.
- If possible, run Drupal on the same node as mariadb (lower weight, first rule should have priority).

I tested this by scaling the deployment to 6 replicas, and they were put on 6 separate nodes. When scaled to 1 replica, that pod is on the same node as the mariadb pod, given that the node has resources available.


Jancis:
What happens when there are less nodes in cluster than requested replicas?

floretan:
I think we discussed this, but just so it's documented: In case there are less nodes than requested replicas some pods will be on the same node.

PR #162, opened 2019-09-30
Feature/backup separation

Jancis:
adds `{{ .Release.Name }}` prefix for backup volume

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -315,6 +315,7 @@ varnish:
 elasticsearch:
   enabled: false
   replicas: 1
+  minimumMasterNodes: 1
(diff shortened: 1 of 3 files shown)


chart/values.yaml
```diff
@@ -315,6 +315,7 @@ varnish:
 elasticsearch:
   enabled: false
   replicas: 1
+  minimumMasterNodes: 1
```
floretan: This doesn't belong in here :-)

Jancis: yep, noticed the difference between chart and k8s, decided to bring it along. it's a separate commit :)

PR #173, opened 2019-10-28
Adds NewRelic APM integration for Drupal

agnis-mateuss:
! **PHP-FPM PR required: https://github.com/wunderio/drupal-php-fpm/pull/15** 

Adds Drupal application reporting to NewRelic. Disabled by default.
Application is labeled as `<KubernetesNamespace>-<branch>` in NewRelic.

### Testing
~1. Set `php.newrelic` as enabled in your silta.yml~ (newrelic is auto-enabled when license is defined)
2. Add secrets file, containing NewRelic API key. This file must define it as `php.newrelic.license`
3. Dont forget to add `decrypt_files` step in `.circle/config.yml` and to pass it to `silta_config`.
4. After deployment, the application shows up few minutes later in NewRelic.

~Notice, the changes use LAGOON_* and NEWRELIC_* environment variables which are parsed into newrelic.ini by the mechanism defined in https://github.com/amazeeio/lagoon/blob/8d9fd7057bab6f151277bcdc0a8aac5f8edaa3a3/images/php/fpm/Dockerfile#L116 .~

phpinfo():
```
newrelic.appname => drupal-project-k8s-feature/SLT-366 => drupal-project-k8s-feature/SLT-366
newrelic.attributes.enabled => enabled => enabled
newrelic.license => ***REDACTED*** => ***REDACTED***
```


chart/templates/_helpers.tpl
```diff
@@ -146,6 +146,17 @@ imagePullSecrets:
       key: hashsalt
 - name: DRUPAL_CONFIG_PATH
   value: {{ .Values.php.drupalConfigPath }}
+{{- if .Values.php.newrelic.enabled }}
```
floretan: If we only need a license, and the default license is empty, can we check for the presence of a license instead of a dedicated `enabled` parameter?

agnis-mateuss: From existence of license key, infer whether newrelic is required. This is an elegant suggestion, Ill add it.

floretan:
I had a look at the newrelic.ini, and I'm wondering if we couldn't modify this file when we extend the base image to use the existing environment variables instead of the ones with the LAGOON_* prefix.

Jancis:
Changes added: 
 - rewriting newrelic entrypoint to string-replace appname with local environment variables;
 - environment variable added: `$PROJECT_NAME`

PR #180, opened 2019-11-13
Feature/silta varnish

Jancis:
This change probes `/robots.txt` file and does `drush sql-drop` before importing dump in post-install hook.

Fresh post-install with https://feature-silta-varnish.drupal-project-k8s.silta.wdr.io
Cleaned and rebuilt this twice in a row.

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -131,6 +131,9 @@ php:
     command: |
       if [ -f reference-data/db.sql.gz ]; then
 
+        echo "Dropping old database"
+        drush sql-drop -y
+
         echo "Importing reference database dump"
         cat reference-data/db.sql.gz | gunzip | drush sql-cli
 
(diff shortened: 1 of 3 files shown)


chart/values.yaml
```diff
@@ -131,6 +131,9 @@ php:
     command: |
       if [ -f reference-data/db.sql.gz ]; then
 
+        echo "Dropping old database"
```
floretan: Is this still needed when requests are sent to /robots.txt?

Jancis: It appears so. I removed it and this happened. 
```
2019-11-15T10:44:32.25444349Z Waiting for database.
2019-11-15T10:44:32.405068539Z Uptime: 26  Threads: 1  Questions: 2  Slow queries: 0  Opens: 17  Flush tables: 1  Open tables: 11  Queries per second avg: 0.076
2019-11-15T10:44:32.45367475Z Importing reference database dump
2019-11-15T10:44:33.076161324Z > ERROR>  1050 (42S01)>  at line 25> : Table 'batch' already exists
Too long with no output (exceeded 10m0s)
```
I believe this only happens when I delete deployment and redeploy from scracth.

Will retry with this re-enabled

Jancis: Yep, works again. Let's keep it then.

PR #181, opened 2019-11-18
SLT-383: consolidate ConfigMaps

floretan:
We currently have 5 different configMaps which are mounted to the drupal pods. This means we get a lot more mounted volumes (which are also watched for changes), and I believe this is what causes us to have pods not provisioning successfully on larger nodes. Aside having one slightly larger file, I don't see a drawback of having different Drupal-related configurations in a single configMap.


Jancis:
Do you think this would be a good use for configmap checksum label too?

floretan:
Fixed the broken conflict resolution and added the checksum.

PR #186, opened 2019-12-12
Allow varnish BAN requests from internal ip's

Jancis:


chart/templates/varnish-configmap-vcl.yaml
```diff
@@ -41,9 +41,12 @@ data:
 
     # Define the purge network access.
     # These are used below to allow cache purging via BAN.
-    # acl purge {
-    #   "127.0.0.1";
-    # }
+    acl purge {
+      "127.0.0.1";
+      {{- range .Values.nginx.noauthips }}
```
floretan: We whitelist the internal IPs from basic authentication, and it makes sense to also use that for IPs that can BAN varnish content, but do we have any reason to also include external IPs here too?

Jancis: yes, these IP addresses can make BAN request to remove stale content (based on cache tags) from cache.
actually, anyone was able to do it currently - https://github.com/wunderio/drupal-project-k8s/pull/186/files#diff-6ca7e79c2d114b9b7ab9d82122fc6fc8R299

Jancis: purge ACL is not used anywhere else right now.

Jancis: which means, once basic auth is removed, there's nothing that prevents anyone from banning pages.
and 127.0.0.1 is not enough since web/cron containers are making BAN requests to varnish container and they are identified as gke network traffic.

PR #191, opened 2020-01-28
SLT-382: Custom nginx redirects

Jancis:
[nginx map directive](http://nginx.org/en/docs/http/ngx_http_map_module.html) is being leveraged for this task. The list of defined redirects is matched with current url and assigned a variable to custom `$redirect_uri` variable in case it does match. Mapping is done outside of `server` scope, so that full url (with protocol and domain name) can be matched too. Then, the `$redirect_uri` variable is used for actual redirection in `server` scope.

There are following test redirects:
```
nginx:
  redirects: |
    /test1 / ;
    /test2 /test2-redirect ;
    http://feature-redirects.drupal-project-k8s.silta.wdr.io/test3 /test3-redirect ;
    '://~feature-redirects.drupal-project-k8s.silta.wdr.io/test4$' /test4-redirect ;
    ~/test5$ /test5-redirect ;
    ~/test6 /test6-redirect ;
    http://feature-redirects-exposedomain.drupal-project-k8s.silta.wdr.io/test7 /test7-redirect ;
```

---

And this is the result for each of them:

 - works, uses absolute path (`/test`, exact matching):
```
➜  ~ curl https://feature-redirects.drupal-project-k8s.silta.wdr.io/test1 -IL
HTTP/1.1 301 Moved Permanently
Location: https://feature-redirects.drupal-project-k8s.silta.wdr.io/
```

 - works, uses absolute path (`/test`, exact matching):
```
➜  ~ curl https://feature-redirects.drupal-project-k8s.silta.wdr.io/test2 -IL 
HTTP/1.1 301 Moved Permanently
Location: https://feature-redirects.drupal-project-k8s.silta.wdr.io/test2-redirect
```

 - works, full url matching
```
➜  ~ curl https://feature-redirects.drupal-project-k8s.silta.wdr.io/test3 -IL
HTTP/1.1 301 Moved Permanently
Location: https://feature-redirects.drupal-project-k8s.silta.wdr.io/test3-redirect
```
 - XX ~works~, full url matching, protocol as wildcard
```
➜  ~ curl https://feature-redirects.drupal-project-k8s.silta.wdr.io/test4 -IL
HTTP/1.1 301 Moved Permanently
Location: https://feature-redirects.drupal-project-k8s.silta.wdr.io/test4-redirect
```

 - Exact path (all exposeDomains)
```
➜  ~ curl https://feature-redirects.drupal-project-k8s.silta.wdr.io/test5 -IL
HTTP/1.1 301 Moved Permanently
Location: https://feature-redirects.drupal-project-k8s.silta.wdr.io/test5-redirect

➜  ~ curl https://feature-redirects.drupal-project-k8s.silta.wdr.io/test5/subpath -IL
HTTP/2 404
```

 - Reacts to all subpaths too:
```
➜  ~ curl https://feature-redirects.drupal-project-k8s.silta.wdr.io/test6 -I
HTTP/1.1 301 Moved Permanently
Location: https://feature-redirects.drupal-project-k8s.silta.wdr.io/test-6-redirect

➜  ~ curl https://feature-redirects.drupal-project-k8s.silta.wdr.io/test6/subpath -I
HTTP/1.1 301 Moved Permanently
Location: https://feature-redirects.drcupal-project-k8s.silta.wdr.io/test-6-redirect
```

 - Only reacts to specific domain
```
➜  ~ curl https://feature-redirects.drupal-project-k8s.silta.wdr.io/test7 -Ik 
HTTP/2 404 

➜  ~ curl https://feature-redirects-exposedomain.drupal-project-k8s.silta.wdr.io/test7 -Ik
HTTP/1.1 301 Moved Permanently
Location: https://feature-redirects-exposedomain.drupal-project-k8s.silta.wdr.io/test7-redirect
```

### DIFF
--- a/chart/templates/drupal-configmap.yaml
+++ b/chart/templates/drupal-configmap.yaml
@@ -237,6 +237,8 @@ data:
         add_header                  Strict-Transport-Security max-age=31536000;
         add_header                  X-XSS-Protection '1; mode=block';
 
+        map_hash_bucket_size 128;
+
         map $uri $no_slash_uri {
             ~^/(?<no_slash>.*)$ $no_slash;
         }
@@ -307,10 +309,21 @@ data:
         https on;
     }
 
+    {{- if .Values.nginx.redirects }}
+    # Custom redirects
+    map '$scheme://$host$request_uri' $redirect_uri {
+        {{- .Values.nginx.redirects | nindent 8 -}}
+    }
+    {{- end }}
 
     server {
         server_name drupal;
         listen 80;
+
+        # Redirects to specified path if map returns anything
+        if ($redirect_uri) {
+    	    return 301 $redirect_uri;
+        }
 
         root /app/web;
         index index.php;
--- a/chart/templates/drupal-configmap.yaml
+++ b/chart/templates/drupal-configmap.yaml
@@ -310,9 +310,21 @@ data:
     }
 
     {{- if .Values.nginx.redirects }}
-    # Custom redirects
+    # Custom redirects with full url matching
     map '$scheme://$host$request_uri' $redirect_uri {
-        {{- .Values.nginx.redirects | nindent 8 -}}
+        {{- range .Values.nginx.redirects }}
+        {{- if contains "://" .from }}
+        {{ .from | squote }} {{ .to | squote }};
+        {{- end }}
+        {{- end }}
+    }
+    # Custom redirects with absolute path matching
+    map $request_uri $redirect_uri_local {
+        {{- range .Values.nginx.redirects }}
+        {{- if not ( contains "://" .from ) }}
+        {{ .from | squote }} {{ .to | squote }};
+        {{- end }}
+        {{- end }}
     }
     {{- end }}
 
@@ -323,6 +335,9 @@ data:
         # Redirects to specified path if map returns anything
         if ($redirect_uri) {
     	    return 301 $redirect_uri;
+        }
+        if ($redirect_uri_local) {
+    	    return 301 $redirect_uri_local;
         }
 
         root /app/web;
--- a/chart/templates/drupal-configmap.yaml
+++ b/chart/templates/drupal-configmap.yaml
@@ -332,6 +332,7 @@ data:
         server_name drupal;
         listen 80;
 
+        {{- if .Values.nginx.redirects }}
         # Redirects to specified path if map returns anything
         if ($redirect_uri) {
     	    return 301 $redirect_uri;
@@ -339,6 +340,7 @@ data:
         if ($redirect_uri_local) {
     	    return 301 $redirect_uri_local;
         }
+        {{- end }}
 
         root /app/web;
         index index.php;
--- a/silta/silta.yml
+++ b/silta/silta.yml
@@ -9,3 +9,19 @@ varnish:
 
 elasticsearch:
   enabled: true
+
+exposeDomains:
+  - hostname: feature-redirects-exposedomain.drupal-project-k8s.silta.wdr.io
+    ssl:
+      enabled: true
+      issuer: letsencrypt-staging
+
+nginx:
+  redirects: |
+    /test1 / ;
+    /test2 /test2-redirect ;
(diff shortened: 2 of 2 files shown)


chart/templates/drupal-configmap.yaml
```diff
@@ -237,6 +237,8 @@ data:
         add_header                  Strict-Transport-Security max-age=31536000;
         add_header                  X-XSS-Protection '1; mode=block';
 
+        map_hash_bucket_size 128;
```
floretan: Why is this needed?

silta/silta.yml
```diff
@@ -9,3 +9,19 @@ varnish:
...
   enabled: true
+
+exposeDomains:
+  - hostname: feature-redirects-exposedomain.drupal-project-k8s.silta.wdr.io
+    ssl:
+      enabled: true
+      issuer: letsencrypt-staging
+
+nginx:
+  redirects: |
```
floretan: If we do some preprocessing on these values, wouldn't it make more sense to list them as a yaml array/object so we can cleanly iterate through them?

chart/templates/drupal-configmap.yaml
Jancis: because nginx didn't start and returned this error: 
> Nginx: could not build map_hash, you should increase map_hash_bucket_size: 64

https://serverfault.com/questions/853102/nginx-could-not-build-map-hash-you-should-increase-map-hash-bucket-size-64
So i set it to the next value or 128 and that helped. 
http://nginx.org/en/docs/http/ngx_http_map_module.html#map_hash_bucket_size

silta/silta.yml
Jancis: Only in case we actually do preprocessing on them. 

If not, i (personally) would be happy to keep it as a string because some really long boring lists of redirects would need a list item prefix `-` then.

I am not against defining a list too.

chart/templates/drupal-configmap.yaml
Jancis: If anyone stumbles accross an error, we can extract the `map_hash_bucket_size` as a variable, but until then, I vote for keeping it simple.

Jancis:
Just adding `-` in front of each redirect and making it a list won't work. Processing entries like this is wrong, because it will also find `://` in destination url.
```
{{- if .Values.nginx.redirects }}
    # Custom redirects with full url matching
    map '$scheme://$host$request_uri' $redirect_uri {
        {{- range .Values.nginx.redirects }}
        {{ if contains "://" . }}
        {{ . }}
        {{- end }}
        {{- end }}
    }
    # Custom redirects with full absolute path matching
    map $request_uri $redirect_uri {
        {{- range .Values.nginx.redirects }}
        {{ if not ( contains "://" . ) }}
        {{ . }}
        {{- end }}
        {{- end }}
    }
{{- end }}
```

To counter that, i added `from` and `to` keys, which changes the definition from 
```
nginx:
  redirects: |
    /test1 / ;
    /test2 /test2-redirect ;
    http://feature-redirects.drupal-project-k8s.silta.wdr.io/test3 /test3-redirect ;
    '~feature-redirects.drupal-project-k8s.silta.wdr.io/test4$' /test4-redirect ;
    ~/test5$ /test5-redirect ;
    ~/test6 /test6-redirect ;
    http://feature-redirects-exposedomain.drupal-project-k8s.silta.wdr.io/test7 /test7-redirect ;
```
to
```
nginx:
  redirects:
    - from: /test1 
      to: /
    - from: /test2 
      to: /test2-redirect
    - from: http://feature-redirects.drupal-project-k8s.silta.wdr.io/test3 
      to: /test3-redirect
    - from: '~feature-redirects.drupal-project-k8s.silta.wdr.io/test4$' 
      to: /test4-redirect
    - from: ~/test5$ 
      to: /test5-redirect
    - from: ~/test6 
      to: /test6-redirect
    - from: http://feature-redirects-exposedomain.drupal-project-k8s.silta.wdr.io/test7 
      to: /test7-redirect
```
The later getting too long and too complex for my liking.

Jancis:
@floretan this is open for review again. I changed the structure of redirect definition (see original post, it's edited) and templates accordingly.
When you accept PR, i'll remove values from silta.yml

PR #197, opened 2020-02-19
Use monolog to send Drupal logs to stdout

floretan:
Looking for a way to get Drupal logs into Stackdriver, this is based on what a few other Kubernetes users in the Drupal community are doing. There is a dependency on the monolog module, but if the module is not present everything works just fine. Since the service modification are injected via a configmap, there is no need for any other modification besides installing and enabling the module.

Here's what the logs look like:
```
[2020-02-19 14:17:32] content.NOTICE: article: updated Bob Ross Ipsum. {"link":"[object] (Drupal\\Core\\GeneratedLink: \"<a href=\\\"/node/3\\\" hreflang=\\\"en\\\">View</a>\")"} {"referer":"https://feature-monolog.drupal-project-k8s.silta.wdr.io/node/3/edit?destination=/admin/content","ip":"88.217.174.125","request_uri":"http://feature-monolog.drupal-project-k8s.silta.wdr.io/node/3/edit?destination=%2Fadmin%2Fcontent","uid":"1","user":"Mrs. Sandrine Kozey"}
```
There is definitely potential for improvement there, but it has the information we need.

For getting this to projects, we need to publish the chart changes, and add the monolog module to all silta projects as well to the drupal-project template.


Jancis:
I tried several attempts to get logs from watchdog/logger to syslog to stderr, but didn't find a good solution that would work without sidecar or extra overhead. Let's go with this one, looks clean and exactly what is needed.

PR #201, opened 2020-03-12
SLT-460: fix the php.ini location

floretan:
There's currently a mismatch between the location where we mount our php.ini configuration, and the expected location from the base image. Things still works though, although these configuration files did nothing. This PR fixes the main file with our overrides, removes additional files that were not needed, and also removes unnecessary overrides of the underlying defaults.

We should also revise the remaining overrides. There are some that we probably want to make configurable (like `memory_limit`), and some where we should consider why they should differ from the defaults and document that.

### DIFF
--- a/chart/templates/drupal-configmap.yaml
+++ b/chart/templates/drupal-configmap.yaml
@@ -154,30 +154,6 @@ data:
     [igbinary]
     igbinary.compact_strings=1
 
-
-  php-fpm_conf: |
-    [global]
-    error_log = /proc/self/fd/2
-    log_level = {{ .Values.php.php_ini.loglevel }} ; Possible Values: alert, error, warning, notice, debug
-    daemonize = no
-    include=/etc/php7/php-fpm.d/*.conf
-
-  www_conf: |
-    [www]
-    listen = [::]:9000
-    user = www-data
-    group = www-data
-    pm = ondemand
-    pm.max_children = 100
-    pm.start_servers = 5
-    pm.min_spare_servers = 5
-    pm.max_spare_servers = 20
-    pm.process_idle_timeout = 60s
-    pm.max_requests = 500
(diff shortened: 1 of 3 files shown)


chart/templates/drupal-configmap.yaml
```diff
@@ -13,171 +13,53 @@ data:
...
-    user = www-data
-    group = www-data
-    pm = ondemand
-    pm.max_children = 100
-    pm.start_servers = 5
-    pm.min_spare_servers = 5
-    pm.max_spare_servers = 20
-    pm.process_idle_timeout = 60s
-    pm.max_requests = 500
-    access.log = /proc/self/fd/2
```
Jancis: We need this override

floretan: These overrides were not used at all for a while. The default values are here: https://github.com/amazeeio/lagoon/blob/db05a8cad8cdb0f36e2d16a462a7ef83e1450f3f/images/php/fpm/php-fpm.d/www.conf, and before we override them we should have a clear reasoning for doing it.

floretan:
I like the idea of using a single text blob for php.ini overrides (or a silta.ini file). The problem comes from using certain specific values like `post_max_size` in multiple places, for example to make sure that the nginx configuration is compatible.

Jancis:
solution a: any chance that php override blob can be a key-value list instead? can still reuse `post_max_size` with a defaults
solution b: have a dedicated value "upload_site" or similar, that overrides relevant variables, no matter what they are set to in the override blob

Jancis:
other than that it's goo to go as - is

floretan:
As discussed on slack: we can keep the current structure with specific configurable keys, and we'll add a free-form ini file in a separate issue. Since ini files have sections, we would need multilevel objects to represent the structure, which would get pretty messy.

PR #205, opened 2020-03-19
Feature/networkpolicy

Jancis:
### DO NOT MERGE, SOME OF THIS WILL BE ADDED TO LIBRARY SUBCHART

Adds a networkpolicy resource to each pod in release that denies connections from all namespaces but `silta-cluster` and the current.

Configuration definition is still open for debate if there are any suggestions (i.e. ingress vs egress if we want to go there)
```
networkPolicy:
  allowNamespaces: ['silta-cluster']
```


floretan:
Integrated into https://github.com/wunderio/drupal-project-k8s/pull/210

PR #209, opened 2020-03-25
Start copying reference data files right away

floretan:
We recently sped things up by importing files in parallel with the reference database dump, but we can already do that before the database and elasticsearch are ready, and we can also continue while the configuration import and other drush commands are executing.


Jancis:
This is a great change, thanks!

PR #214, opened 2020-04-08
Introducing silta-ingress: allow

Jancis:
Introduces new label for pods that allow traffic from `silta-cluster` namespace.
Web and shell access works.

### DIFF
--- a/charts/drupal/templates/network-policy.yaml
+++ b/charts/drupal/templates/network-policy.yaml
@@ -1,6 +1,0 @@
-# Allow access to drupal or varnish pods from the ingress in the silta-cluster namespace
-{{- $id := "allow-ingress" -}}
-{{- $podSelectors := dict "silta-ingress" "allow" -}}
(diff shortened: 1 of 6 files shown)


charts/drupal/templates/network-policy.yaml
```diff
@@ -1,6 +1,6 @@
 # Allow access to drupal or varnish pods from the ingress in the silta-cluster namespace
 {{- $id := "allow-ingress" -}}
-{{- $podSelectors := ternary (dict "service" "varnish") (dict "app" "drupal") .Values.varnish.enabled -}}
+{{- $podSelectors := dict "silta-ingress" "allow" -}}
```
floretan, line 3: This needed to be in the drupal chart before because `.Values.varnish.enabled` is not available in the subchart. Now that it's not needed anymore, we can move this whole file to the silta-release chart.

Jancis: Drupal networkpolicy moved to silta-release subchart networkpolicy as a dedicated section.

PR #215, opened 2020-04-09
Feature/np context fix

Jancis:


Jancis:
Test np's are generated correctly now :tada:

PR #220, opened 2020-04-29
Feature/slt 477

agnis-mateuss:
Enabling of Newrelic plugin triggers on presence of NEWRELIC_ENABLED env variable during build, so we just remove this variable. No changes needed to base docker image in this way.


Jancis:
Did you also change our upstream image that uses the env variables?

PR #221, opened 2020-05-04
SLT-485: Disable Instana Twig instrumentation

floretan:
The way Drupal uses Twig can cause an excessive number of independent calls to Instana which generates a high CPU activity on our servers.


Jancis:
Can i suggest adding a link to instana documentation as the case is quite specific and the instrumentation id sounds so arbitrary?

PR #234, opened 2020-06-08
Feature/slt 318

agnis-mateuss:
Restores database and files from a backup, given a date.
Refactored post-release-command in _helpers.tpl to separate drupal installation + followup deployments from backup deployments.
Introduces postRestoreCommand variable in values.yaml to specify a command to execute after backup is imported.

Eventual merge is dependant on https://github.com/wunderio/silta-circleci/pull/100 being merged first.

Testing:
1. Create an environment, and backup
2. Touch some files in public-files, add articles in drupal.
3. Change CircleCI config.yml to use `backup` branch -> `silta/silta@dev:backup`
3. Push a commit with message #DestroyCurrentAndRestoreBackup=[yyyy-mm-dd]-[branch], date of the backup made.
4. Check that another backup is made - to save the current state, before the targeted backup is imported.
5. Extra files from step 2 are removed. Drupal articles should be gone.


charts/drupal/templates/_helpers.tpl
```diff
@@ -235,40 +235,67 @@ done
...
+
+  # Restore db from targeted backup
+  {{ include "drupal.import-backup-db" . }}
+
+  {{ if .Values.elasticsearch.enabled }}
+    {{ include "drupal.wait-for-elasticsearch-command" . }}
+  {{ end }}
+
+  # Running custom commands after restored backup
+  {{ .Values.php.postRestoreCommand }}
```
floretan, line 296: At this point we still need to run `{{ .Values.php.postupgrade.command}}`, for example to run database updates.

agnis-mateuss: Added right before executing postRestoreCommand

PR #252, opened 2020-09-09
Prefer keeping pods in the same zone.

floretan:
This makes no difference for single-zone clusters, but we do get a small benefit with regional clusters.


Jancis:
Sounds like a good idea!

PR #257, opened 2020-09-30
ingress aggregation

Jancis:
This PR:
  - aggregates exposeDomains into ingresses (no more index per exposeDomain)
  - Adds ingress definition (and deprecates `exposeDomains.domain.type`)
  - Promotes map definition instead of list for exposeDomains
  - Adds test cases for things above


agnis-mateuss:
Here it goes!

agnis-mateuss (approved):

PR #268, opened 2020-10-21
Feature/slt 505 enable memcache settingsphp

mgalang:
Description

User story: https://wunder.atlassian.net/browse/SLT-505

Changes to charts: https://github.com/wunderio/charts/pull/160

Set the default cache backend to use memcache when a memcache server is available and when memcache cache backend is defined.

At the moment we only set the memcache host server settings, but it remains unused because it is not set as the backend cache for Drupal.


Jancis (approved):
Awesome!

PR #273, opened 2020-10-30
Drupal 7 aware drush commands

Jancis:
Since Drupal 7 projects are using this chart now, the build process might fail when `drush cr` is executed.
Tried several ways to detect major version, even define via `.Values.drupal.majorVersion`, but this seems like the least intrusive one.


Rade333 (approved):
Looks good :+1:

Rade333 (approved):

PR #275, opened 2020-11-11
SLT-505: Port the memcache settings changes from charts repository

mgalang:
### Description

Updated the memcache setting changes from the charts repository.

### DIFF
--- a/charts/drupal/files/settings.silta.php
+++ b/charts/drupal/files/settings.silta.php
@@ -5,6 +5,8 @@
  * Automatically injected settings for the Silta environment.
  */
 
+use Drupal\Core\Installer\InstallerKernel;
+
 // Database settings.
 $databases['default']['default'] = [
   'database' => getenv('DB_NAME'),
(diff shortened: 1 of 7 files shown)


charts/drupal/files/settings.silta.php
```diff
@@ -5,6 +5,8 @@
  * Automatically injected settings for the Silta environment.
  */
 
+use Drupal\Core\Installer\InstallerKernel;
```
Jancis, line 8: 1. Does this work with D8?
2. Will there be any performance penalties because of this include?

mgalang: @Jancis This is specifically for D8. As far as I know, no performance penalties.

Jancis: Whoops, meant D7. We'll need another file for D7.

Jancis:
This might break D7 deployments, so let's make a slight change where developers can define/override drupal version in values and have multiple settings.php files.

1. Because helm does not know the drupal version, we'll need to have configuration parameter for it. i.e. this -
```
drupal:
  coreVersion: 9
```
2. Make `files/settings7.silta.php` and rename `files/settings.silta.php` to `files/settings8.silta.php`
3. Mount either `files/settings8.silta.php` or `files/settings7.silta.php` based on `drupal.coreVersion` on `sites/default/settings.silta.php`

Jancis (approved):
Great collaboration, looks good to me now!

PR #282, opened 2020-11-27
Feature/solr

Jancis:
This implements Solr functionality for drupal chart using [geerlingguy's solr docker images ](https://github.com/geerlingguy/solr-container)(official images support 5+, while we have projects using solr 4).
This is embedded into drupal chart so we can reuse php image to copy solr schema from repo folder into solr container using initContainer. All of that to avoid extra image build (this might change in a future).


agnis-mateuss (approved):
Looks solid!

PR #312, opened 2021-03-23
Removing D8 settings, adopting memcache and varnish settings

ArtisKrumins:


Jancis (approved):
Thanks for finetuning D7 settings!

PR #313, opened 2021-03-24
Drupal chart: Fix failing postupgrade drush commands

Rade333:
Changes proposed in this PR:
- reorder the default postupgrade commands in Drupal chart so that cache rebuild is the very first command being run
- apparently my IDE removed a bunch of whitespaces as well

So before the order of commands was:

1. drush status
2. drush cache-clear all OR drush cache-rebuild
3. drush updatedb
4. drush config-import
5. drush cache-rebuild

After this PR, the order will be:

1. drush cache-clear all OR drush cache-rebuild
2. drush status
3. drush updatedb
4. drush config-import
5. drush cache-rebuild

Link to Slack discussion: https://wunder.slack.com/archives/C8UN6AG9W/p1616578842088100

### DIFF
--- a/charts/drupal/values.yaml
+++ b/charts/drupal/values.yaml
@@ -57,7 +57,7 @@ autoscaling:
 #     hostname: www.example.com
 #   example_no_https:
 #    hostname: insecure.example.com
-#      ssl: 
+#      ssl:
 #        enabled: false
 exposeDomains: {}
 
@@ -79,7 +79,7 @@ ingress:
     # staticIpAddressName: null
 
 # Infrastructure related settings.
-cluster: 
+cluster:
   type: gke
   vpcNative: false
 
@@ -114,7 +114,7 @@ nginx:
     requests:
       cpu: 1m
       memory: 10M
-  
+
   # Set of values to enable and use http basic authentication
   # It is implemented only for very basic protection (like web crawlers)
   basicauth:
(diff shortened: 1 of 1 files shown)


Jancis (changes requested):
drush `cache-rebuild` fails when the bootstrap can't be executed
> Command cache-rebuild needs a higher bootstrap level to run - you    [error]
> will need to invoke drush from a more functional Drupal environment
> to run this command.

This means a repeated deployment will fail and You won't be able to have successful deployment until You import the database (which might not be possible if shell was not enabled on the first run (postinstall vs postupgrade).

Based on this, the change can't be accepted, unfortunately.

I would be very happy to accept the whitespace fixes, though :)

Rade333:
> drush `cache-rebuild` fails when the bootstrap can't be executed
> 
> > Command cache-rebuild needs a higher bootstrap level to run - you    [error]
> > will need to invoke drush from a more functional Drupal environment
> > to run this command.
> 
> This means a repeated deployment will fail and You won't be able to have successful deployment until You import the database (which might not be possible if shell was not enabled on the first run (postinstall vs postupgrade).
> 
> Based on this, the change can't be accepted, unfortunately.
> 
> I would be very happy to accept the whitespace fixes, though :)

Thanks for the feedback!

I did notice that drush cr does _not_ fail when run on an empty database. To my understanding, when the deploy process reaches the postupgrade commands, the database already exists (even though it may be empty), and the credentials should be in order for drush to be able to connect to it, right? So in that case this shouldn't be a problem, but maybe I'm missing something here?

Jancis:
You're correct, I was testing without database. Drush status returns with exit code 0 when the database is there, but empty, so this should be fine after all. Thank You for re-checking it!

Jancis (approved):

PR #319, opened 2021-05-31
Make varnish more configurable

badrange:
When we try to ddos our silta-prod environment we see some 503 errors from Varnish and we'd like to see if we can avoid those errors by adjusting connect_timeout (at least). 

I figured it could be nice to make the other probe settings configurable while I was at it.

### DIFF
--- a/charts/drupal/templates/varnish-configmap-vcl.yaml
+++ b/charts/drupal/templates/varnish-configmap-vcl.yaml
@@ -15,19 +15,19 @@ data:
     backend prod1_http1 {
       .host = "{{ .Release.Name }}-drupal";
       .port = "80";
-      .max_connections = 300;
+      .max_connections = {{ int .Values.varnish.max_connections }};
 
       .probe = {
-        .url       = "/robots.txt"; 
-        .interval  = 5s; # check the health of each backend every n seconds
-        .timeout   = 1s; # timing out after n seconds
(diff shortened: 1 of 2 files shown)


Jancis:
Since there could be a bit more parameters in the [varnish probe definition](https://varnish-cache.org/docs/6.0/reference/vcl.html#probes) than defined currently (and they might still be needed), i'd suggest extracting the whole probe section as plaintext.
I always dislike the variable per line situations too.
So I'd extract the probes definition to a string that defaults to existing one, and then have another for backend, except .host and .port. Probably, it would all just be a backend config.
something like this
```
varnish:
  vcl_extra_cookies: ""
  # Do not cache files larger than 3 megabytes (use integers).
  cache_skip_size: 3
  # Backend configuration (except host and port)
  # https://varnish-cache.org/docs/6.0/reference/vcl.html#backend-definition
  backend_config: |
    .max_connections = 300;
      .probe = {
        .url       = "/robots.txt"; 
        .interval  = 5s;
        (..)
      }
      (..)
```

By the way, default the backend.probe.timeout to 5s already since it looks like more sensible value for drupal projects.
And, put host and port below the include part. Not sure which takes precedence (or if varnishd even allows redefining the same keys), but just in case someone tries to do something fishy, chart default should be defined after the customized part.

Jancis:
Closed in favor of https://github.com/wunderio/drupal-project-k8s/pull/320

PR #320, opened 2021-05-31
Make varnish backend config configurable

badrange:
Bonus: Increase default varnish probe timeout to 5s to hopefully avoid 503s under high load.

### DIFF
--- a/charts/drupal/templates/varnish-configmap-vcl.yaml
+++ b/charts/drupal/templates/varnish-configmap-vcl.yaml
@@ -15,7 +15,7 @@ data:
     backend prod1_http1 {
       .host = "{{ .Release.Name }}-drupal";
       .port = "80";
-      {{ .Values.varnish.backend_config }}
+      {{- .Values.varnish.backend_config | nindent 6 }}
     }
  
     # Define the internal network access.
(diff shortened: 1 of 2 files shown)


charts/drupal/templates/varnish-configmap-vcl.yaml
```diff
@@ -15,19 +15,7 @@ data:
     backend prod1_http1 {
       .host = "{{ .Release.Name }}-drupal";
       .port = "80";
-      .max_connections = 300;
-
-      .probe = {
-        .url       = "/robots.txt"; 
-        .interval  = 5s; # check the health of each backend every n seconds
```
Jancis, line 22: Let's remove those doc comments. Variable names are self explanatory on the most part and there's also a link to VCL file at the top. It will make the yaml file look slightly cleaner, less busy and more readable.

charts/drupal/templates/varnish-configmap-vcl.yaml
```diff
@@ -15,19 +15,7 @@ data:
...
-        .timeout   = 1s; # timing out after n seconds
```
Jancis, line 23: Let's increase this to 5s, in case of slow response of a single page, it might bring the whole backend offline. And there's only a single backend contact point (service) in this architecture.

badrange: Done, in `charts/drupal/values.yaml` line 466

charts/drupal/templates/varnish-configmap-vcl.yaml
badrange: Makes sense! Done (in `charts/drupal/values.yaml`)

Jancis (approved):
Excellent work, thank You very much for this addition!

PR #324, opened 2021-06-08
Make Drupal installation path configurable.

tharna:
When running production on 3rd party hosting for example in Acquia cloud the Drupal installation path needs to be different from our standard `/web`.
We could move the `docroot` path under `/app/web` when building the image, but there are also dependencies on that folder e.g. in `composer.json`.

Related PR: https://github.com/wunderio/silta-circleci/pull/137

### DIFF
--- a/charts/drupal/templates/_helpers.tpl
+++ b/charts/drupal/templates/_helpers.tpl
@@ -35,11 +35,11 @@ ports:
   readOnly: true
   subPath: php_ini
 - name: config
-  mountPath: /app/web/sites/default/settings.silta.php
+  mountPath: /app/{{ .Values.appPath }}/sites/default/settings.silta.php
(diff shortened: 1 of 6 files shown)


Jancis (changes requested):
The `webRoot` inline comments are a bit too excessive, those should be moved into documentation. Especially since it explicitly mentions a certain CI tool. Leave first 1-2 lines of it, the rest should be moved/removed.

Rade333:
> The `webRoot` inline comments are a bit too excessive, those should be moved into documentation. Especially since it explicitly mentions a certain CI tool. Leave first 1-2 lines of it, the rest should be moved/removed.

Good point. Moved docs now and added link to new doc page.

Jancis (approved):

PR #326, opened 2021-06-16
Varnish resource assignment tweaks

Jancis:
Increasing varnish resource requests to reflect a real world deployment and setting limits to avoid shared resource hogging


Jancis:
The resource over-usage might have happened due to broken cache file storage that was fixed here:
https://github.com/wunderio/silta-images/commit/b80789a025dc364256e0d8b113f2074ceecccd53#diff-b7316498cc904e91547389c7df43c6af055b272929ad7cc207d7cf4b14b720f7R11
Closing PR for now.

PR #327, opened 2021-06-17
Increase worker_connections and worker_rlimit_nofile.

tharna:
Added suggested mitigations # 1 and # 4 from https://hexadix.com/slowloris-dos-attack-mitigation-nginx-web-server/
Other limits were already appropriate.


Jancis (approved):
Good find! There will be [another change](https://github.com/wunderio/drupal-project-k8s/pull/328) slightly up in the connection chain, in traefik, that limits connections per ip. These two should tackle the slowloris issue.

PR #329, opened 2021-06-29
Autoscaling defaults for memory

ArtisKrumins:


Jancis (approved):
Excellent addition, thank You!

PR #334, opened 2021-07-09
SLT-611: X-Robots-Tag

ArtisKrumins:


Jancis (approved):

gatis (approved):
looks good

Jancis:
if a nginx proxy tries to load page from default domain, they can do `proxy_hide_header X-Robots-Tag` to allow robots on it.
http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_hide_header

PR #397, opened 2021-11-22
Improve documentation for uploading files. Reason: rsync and "No such file or directory"

ragnarkurmwunder:
https://wunder.slack.com/archives/C8UN6AG9W/p1637562569252600

**The problem**

People try to use Silta instructions for uploading files but fail.

```
rsync -azv --temp-dir=/tmp/ -e 'ssh -A -J www-admin@ssh.dev.wdr.io' files/ www-admin@master-shell.client-fi-turku-kasiohjelma:/app/web/sites/default/files

rsync: [receiver] rename "/tmp//view_unpublished-8.x-1.0-alpha1.fi.po.FonBkJ" -> "translations/view_unpublished-8.x-1.0-alpha1.fi.po": No such file or directory (2)
```

This is due to Google Filestore architecture where dirs do not actually exist. (See longer description in the thread 10:56).

**The solution**

The "quick fix" is that we update the documentation on how to copy files without running into problems.
This is a bit ugly, but at least people won't keep stumbling on this regularly and bugging the Silta team for support.

### DIFF
--- a/charts/drupal/templates/NOTES.txt
+++ b/charts/drupal/templates/NOTES.txt
@@ -62,7 +62,11 @@ Importing data into server (use this with caution!)
   {{ range $index, $mount := .Values.mounts -}}
   {{ if eq $mount.enabled true -}}
   Uploading files to {{ $index }}:
-  rsync -azv --temp-dir=/tmp/ -e 'ssh -A -J {{ include "drupal.jumphost" $ }}' local_folder/ {{ include "drupal.shellHost" $ }}:{{ $mount.mountPath }}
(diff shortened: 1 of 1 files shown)


tormi (approved):

tormi:
This worked flawlessly. Thanks, @ragnarkurmwunder !

Jancis:
@ragnarkurmwunder would there be any way to pass the cp command into ssh? Rarely anyone will copy paste more than one command.
Our storage backend uses Cloud storage API instead of Filestore, but the note itself is redundant, please remove it.

This PR will become redundant once `https://github.com/rclone/rclone/pull/5323` is finally merged and we've deployed it to our clusters.

Jancis:
The rclone PR mentioned above was still not merged so we built our own binary with patch applied to it and are running it on our clusters. Being able to create empty folders should solve the issue, so this workaround is not required anymore.

PR #402, opened 2021-11-29
Feature/slt 665 tls length

agnis-mateuss:
Moves prefix logic into helpers, truncates+hashes environment names if full domain doesnt fit.


Jancis (approved):
Looks alright, just remove the issuer configuration from settings

PR #403, opened 2021-11-30
Backup compression

agnis-mateuss:
Lower compression level in backup pods.


Jancis (approved):
Looks correct, just remove the backup setting (or target to a specfic branch)

PR #468, opened 2022-04-25
Feature/red test

Jancis:
- Initial redis subchart integration
- Chart name validation (will fail if contains one of enabled subcharts like "mailhog", "memcache" or "redis"). This is a special quirk for resource naming in bitnami subcharts https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_names.tpl#L26

### DIFF
--- a/charts/drupal/values.yaml
+++ b/charts/drupal/values.yaml
@@ -575,6 +575,22 @@ memcached:
     # MaxItemSize
     - -I 4M
 
+# https://github.com/bitnami/charts/blob/master/bitnami/redis/values.yaml
+redis:
+  architecture: standalone
+  auth:
+    # mandatory value
+    password: ""
+  replica:
+    # replicaCount: 1
+    resources:
+      limits: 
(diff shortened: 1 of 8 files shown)


guncha25 (approved):
LGTM

charts/drupal/values.yaml
```diff
@@ -575,6 +575,22 @@ memcached:
     # MaxItemSize
     - -I 4M
 
+# https://github.com/bitnami/charts/blob/master/bitnami/redis/values.yaml
+redis:
+  architecture: standalone
```
Rade333, line 580: Should we add the `enabled: false` line here?

Jancis: Good idea, fixed!

Jancis: and then i set it to `enabled: true` :facepalm:

Jancis: false alarm, it was true only in chart tests

PR #471, opened 2022-04-27
Print mariadb disc usage before db service boot

Jancis:
Helps debugging cases when db crashes due to full storage.

### DIFF
--- a/charts/drupal/values.yaml
+++ b/charts/drupal/values.yaml
@@ -470,6 +470,15 @@ mariadb:
         memory: 384M
       limits:
         memory: 512M
+    extraInitContainers: |
+      - name: dbinfo
+        image: busybox:stable
+        command: ["/bin/sh", "-c"]
+        args:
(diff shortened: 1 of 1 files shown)


ArtisKrumins (approved):

charts/drupal/values.yaml
```diff
@@ -470,6 +470,15 @@ mariadb:
         memory: 384M
       limits:
         memory: 512M
+    extraInitContainers: |
+      - name: dbinfo
+        image: busybox:stable
+        command: ["/bin/sh", "-c"]
+        args:
+          - df -h /bitnami/mariadb;
```
ragnarkurmwunder, line 478: Is the semicolon at the end of the line needed?

ragnarkurmwunder (changes requested):

charts/drupal/values.yaml
Jancis: can remove it as well

PR #524, opened 2022-08-15
Improve file backup job to ignore tar warnings

Rade333:
**Changes proposed:**
- Append the drupal backup tar command in order to ignore tar exit code 1

**Reasoning:**

I noticed the backup cron job to fail multiple times, and the last line in logs was `backup tar: /app/web/sites/default/files: file changed as we read it`.

If any files change while tar is creating an archive, it will exit with code 1 and the rest of the backup job will not get run.
From [tar docs](https://www.gnu.org/software/tar/manual/html_section/Synopsis.html): 

> If tar was given ‘--create’, ‘--append’ or ‘--update’ option, this exit code [1] means that some files were changed while being archived and so the resulting archive does not contain the exact copy of the file set.

Changes are based on https://stackoverflow.com/a/49112839/12416374

### DIFF
--- a/charts/drupal/templates/_helpers.tpl
+++ b/charts/drupal/templates/_helpers.tpl
@@ -488,7 +488,7 @@ fi
   # File backup for {{ $index }} volume.
   # If files get changed while the tar command is running, tar will exit with code 1. We ignore this as we want the rest of the job to still get run.
   echo "Starting {{ $index }} volume backup."
-  tar -czP --exclude=css --exclude=js --exclude=styles -f $BACKUP_LOCATION/{{ $index }}.tar.gz {{ $mount.mountPath }}  || ( export exitcode=$?; [[ $exitcode -eq 1 ]] || exit
+  tar -czP --exclude=css --exclude=js --exclude=styles -f $BACKUP_LOCATION/{{ $index }}.tar.gz {{ $mount.mountPath }} || ( export exitcode=$?; [[ $exitcode -eq 1 ]] || exit )
   {{- end -}}
   {{- end }}
   {{- end }}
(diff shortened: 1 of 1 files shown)


Jancis:
Would adding `--warning=no-file-changed` help in any way?
https://www.gnu.org/software/tar/manual/html_section/warnings.html

Rade333:
> Would adding `--warning=no-file-changed` help in any way? https://www.gnu.org/software/tar/manual/html_section/warnings.html

@Jancis unfortunately no. That flag only affects the verbosity, so tar will not output the warning, but it does not affect the exit code.

Jancis:
Looks good, just needs couple minor fixes - there is a double-space before first "or" and the second or with exit is missing a closing/wrapping parenthesis.

Jancis (changes requested):
Looks good, just needs couple minor fixes - there is a double-space before first "or" and the second or with exit is missing a closing/wrapping parenthesis.

Rade333:
> Looks good, just needs couple minor fixes - there is a double-space before first "or" and the second or with exit is missing a closing/wrapping parenthesis.

Good catch! Fixed!

Jancis (approved):

Rade333:
Thank @Jancis for approval, however I'm not authorized to merge :)

Jancis:
The dev workflow still needs some improvement. It will be merged a[nd credited](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) for the next chart release if that's alright.

Rade333:
Sure thing!

Jancis:
I tested backup changes in a dedicated branch, no regressions were noticed.

PR #528, opened 2022-08-31
Make `pv` output visible when importing reference database.

ragnarkurmwunder:
# The problem

Database import takes an hour or more. But I as a user don't know if it is 5min or more.

# The solution

`pv` command is already in the code, but it is useless, as it does not provide any output because it is not running on an active terminal. The `-f` flag forces `pv` to produce output, and that could be seen in the CircleCI task while importing the database.

# Testing

I don't know how to test this, but I assume Silta guys know how.

### DIFF
--- a/charts/drupal/templates/_helpers.tpl
+++ b/charts/drupal/templates/_helpers.tpl
@@ -416,7 +416,7 @@ if [ -f /app/reference-data/db.sql.gz ]; then
 
   echo "Importing reference database dump"
   gunzip -c /app/reference-data/db.sql.gz > /tmp/reference-data-db.sql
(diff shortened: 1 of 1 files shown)


Rade333:
It is at least visible here: https://app.circleci.com/pipelines/github/wunderio/drupal-project-k8s/4229/workflows/b5bf4018-ddb2-4e7a-9f2f-cd1d85fef3df/jobs/14408?invite=true#step-125-51

Looking good 👍 

build-deploy-1 (14408) - wunderio:drupal-project-k8s 2022-09-01 09-14-49

Jancis (approved):

PR #533, opened 2022-09-14
Moving down d7 image derivative generation in nginx location list

Jancis:


TomiMikola (approved):
👏🏻 Thanks for reacting so quickly!

agnis-mateuss (approved):

PR #554, opened 2022-10-17
Faster way of exporting and importing reference database for Drupal chart

Rade333:
Changes proposed:
- Update drupal.extract-reference-data so that it exports one .sql file per database table, and saves them in db.tar.gz
- Update drupal.import-reference-db so that it supports importing the database from db.tar.gz, using xargs to run 10 simultaneous mysql commands, making the import roughly 60% faster than previously
- Keep backwards compability so that the older way of importing db.sql.gz still works
- Improve bash code style

**Note: The new way of importing reference database might consume more memory, so if pods get OOMKilled while importing, the mariadb memory limit needs to be increased.**

**Note 2: Any scripts or commands that developers have used for fetching the reference database manually will need to be updated.**

### DIFF
--- a/charts/drupal/templates/_helpers.tpl
+++ b/charts/drupal/templates/_helpers.tpl
@@ -376,19 +376,45 @@ if [[ "$(drush status --fields=bootstrap)" = *'Successful'* ]] ; then
   echo "Dump reference database."
   dump_dir=/tmp/reference-data-export/
   mkdir "${dump_dir}"
-  for table in $(drush sql-query 'show tables;');
-  do
-    # Add --no-data parameter for tables that match .Values.referenceData.ignoreTableContent.
-    [[ "${table}" =~ {{ .Values.referenceData.ignoreTableContent }} ]] && nodata='--no-data' || nodata=''
-    echo "Dumping table: ${table}"
-    # The ${nodata} variable cannot be quoted in the mysqldump command because if it's empty, the command will fail.
-    mysqldump --user="${DB_USER}" --password="${DB_PASS}" --host="${DB_HOST}" ${nodata} "${DB_NAME}" "${table}" > "${dump_dir}${table}.sql"
+
+  # Figure out which tables to skip the data for.
+  ignore_tables=''
+  for table in $(drush sql-query 'show tables;' | grep -E '{{ .Values.referenceData.ignoreTableContent }}'); do
+    ignore_tables="${ignore_tables} --ignore-table-data=${DB_NAME}.${table}";
   done
+
+  echo "Dump reference database."
+  # The $ignore_tables variable cannot be quoted in the mysqldump command because if it's empty, the command will fail.
+  mysqldump --user="${DB_USER}" --password="${DB_PASS}" --host="${DB_HOST}" ${ignore_tables} "${DB_NAME}" > /tmp/db.sql
+
+  previous_wd=$(pwd)
+  cd "${dump_dir}" || exit
+
+  # Split the dump to one file per table
+  csplit --silent --prefix='table-' /tmp/db.sql '/-- Table structure for table/'-1 {*}
+  # First file is the mysqldump header, rename it to "header"
+  mv table-00 header
+  # Find last table file
+  last_table=$(find -type f -name 'table-*' | sort -n | tail -n1)
+  # Split last table file to extract mysqldump footer, which starts with a line including "@OLD_"
+  csplit --silent --prefix='last-' "${last_table}" '/@OLD_/'
+  # Replace $last_table with the version of it that has footer extracted from it
+  mv last-00 "${last_table}"
+  # Rename the extracted footer to "footer"
+  mv last-01 footer
+  # Prepend header and append footer to all table files, save them as .sql
+  for file in table-*; do
+    table_name=$(grep 'Table structure for table' ${file} | cut -d$'\x60' -f2)
+    cat header "${file}" footer > "${table_name}.sql"
+  done
+  # Remove all non .sql files
+  find . -type f ! -name '*.sql' -delete
+
+  cd "${previous_wd}"
 
   # Compress the sql files into a single file and copy it into the backup folder.
   # We don't do this directly on the volume mount to avoid sending the uncompressed dump across the network.
-  cd "${dump_dir}" || exit
-  tar -cf /tmp/db.tar.gz -I 'gzip -1' ./*.sql
+  tar -cf /tmp/db.tar.gz -I 'gzip -1' -C "${dump_dir}" .
   cp /tmp/db.tar.gz /app/reference-data/db.tar.gz
 
   {{ range $index, $mount := .Values.mounts -}}
--- a/charts/drupal/templates/_helpers.tpl
+++ b/charts/drupal/templates/_helpers.tpl
@@ -373,23 +373,21 @@ done
 {{- define "drupal.extract-reference-data" -}}
 set -e
 if [[ "$(drush status --fields=bootstrap)" = *'Successful'* ]] ; then
-  # Figure out which tables to skip.
-  IGNORE_TABLES=""
-  IGNORED_TABLES=""
-  for TABLE in `drush sql-query "show tables;" | grep -E '{{ .Values.referenceData.ignoreTableContent }}'` ;
+  echo "Dump reference database."
+  mkdir /tmp/reference-data/
+  for TABLE in `drush sql-query "show tables;"`;
   do
-    IGNORE_TABLES="$IGNORE_TABLES --ignore-table=$DB_NAME.$TABLE";
-    IGNORED_TABLES="$IGNORED_TABLES $TABLE";
+    # Add --no-data parameter for tables that match .Values.referenceData.ignoreTableContent.
+    [[ "$TABLE" =~ {{ .Values.referenceData.ignoreTableContent }} ]] && NODATA='--no-data' || NODATA=''
+    echo "Dumping table: $TABLE"
+    mysqldump -u $DB_USER --password=$DB_PASS --host=$DB_HOST $NODATA $DB_NAME $TABLE > /tmp/reference-data/$TABLE.sql
   done
 
-  echo "Dump reference database."
-  mysqldump -u $DB_USER --password=$DB_PASS --host=$DB_HOST $IGNORE_TABLES $DB_NAME > /tmp/db.sql
-  mysqldump -u $DB_USER --password=$DB_PASS --host=$DB_HOST --no-data $DB_NAME $IGNORED_TABLES >> /tmp/db.sql
-
-  # Compress the database dump and copy it into the backup folder.
+  # Compress the sql files into a single file and copy it into the backup folder.
   # We don't do this directly on the volume mount to avoid sending the uncompressed dump across the network.
-  gzip -1 /tmp/db.sql
-  cp /tmp/db.sql.gz /app/reference-data/db.sql.gz
+  cd /tmp/reference-data/
+  tar -cf /tmp/db.tar.gz -I 'gzip -1' *.sql
+  cp /tmp/db.tar.gz /app/reference-data/db.tar.gz
 
   {{ range $index, $mount := .Values.mounts -}}
   {{- if eq $mount.enabled true -}}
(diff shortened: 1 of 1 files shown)


charts/drupal/templates/_helpers.tpl
```diff
@@ -373,23 +373,21 @@ done
 {{- define "drupal.extract-reference-data" -}}
 set -e
 if [[ "$(drush status --fields=bootstrap)" = *'Successful'* ]] ; then
-  # Figure out which tables to skip.
-  IGNORE_TABLES=""
-  IGNORED_TABLES=""
-  for TABLE in `drush sql-query "show tables;" | grep -E '{{ .Values.referenceData.ignoreTableContent }}'` ;
+  echo "Dump reference database."
+  mkdir /tmp/reference-data/
```
ragnarkurmwunder, line 377: any change this dir already exists?
if yes, then the contents may become mixed, if table is removed from db.
if yes, then mkdir fails, because the dir already exists, or it needs to have `-p`.
if no, then all good.

charts/drupal/templates/_helpers.tpl
```diff
@@ -373,23 +373,21 @@ done
...
+  for table in $(drush sql-query 'show tables;');
   do
-    IGNORE_TABLES="$IGNORE_TABLES --ignore-table=$DB_NAME.$TABLE";
-    IGNORED_TABLES="$IGNORED_TABLES $TABLE";
+    # Add --no-data parameter for tables that match .Values.referenceData.ignoreTableContent.
+    [[ "${table}" =~ {{ .Values.referenceData.ignoreTableContent }} ]] && nodata='--no-data' || nodata=''
+    echo "Dumping table: ${table}"
+    mysqldump --user="${DB_USER}" --password="${DB_PASS}" --host="${DB_HOST}" "${nodata}" "${DB_NAME}" "${table}" > /tmp/reference-data/"${table}".sql
```
ragnarkurmwunder, line 383: i see a problem here that the dumps may become inconsistent.
cant lock db during individual table dumping process.
even if this works initially, there will be case where people become confused due to strange problems they witness.
it will be quite hard to debug. it would take silta support resources, and generate dissatisfaction.
there should be some kind of a workaround.

i still think that the general idea is good - parallel import.

we may have a zoom session about options how to do it safely/consistently/atomically.

I'm currently putting `request changes` for this particular item.
to make sure we wont backstab ourselves.

charts/drupal/templates/_helpers.tpl
```diff
@@ -373,23 +373,21 @@ done
...
-  mysqldump -u $DB_USER --password=$DB_PASS --host=$DB_HOST $IGNORE_TABLES $DB_NAME > /tmp/db.sql
-  mysqldump -u $DB_USER --password=$DB_PASS --host=$DB_HOST --no-data $DB_NAME $IGNORED_TABLES >> /tmp/db.sql
-
-  # Compress the database dump and copy it into the backup folder.
+  # Compress the sql files into a single file and copy it into the backup folder.
   # We don't do this directly on the volume mount to avoid sending the uncompressed dump across the network.
-  gzip -1 /tmp/db.sql
-  cp /tmp/db.sql.gz /app/reference-data/db.sql.gz
+  cd /tmp/reference-data/ || exit
+  tar -cf /tmp/db.tar.gz -I 'gzip -1' ./*.sql
```
ragnarkurmwunder, line 389: check out `tar -C`, cd becomes irrelevant on the previous line.
https://linux.die.net/man/1/tar

aha, I see you are already using it below while untarring.
maybe you needed explicit exit when the dir is not there?

charts/drupal/templates/_helpers.tpl
```diff
@@ -412,16 +410,28 @@ fi
...
+  # New way of importing.
+  if [[ -f "${app_ref_data}"/db.tar.gz ]]; then
+    mkdir "${tmp_ref_data}"
+    tar -xzf "${app_ref_data}"/db.tar.gz -C "${tmp_ref_data}"/
+    find "${tmp_ref_data}"/ -type f -name "*.sql" | xargs -P10 -I{} sh -c 'echo "Importing {}" && mysql -A --user="${DB_USER}" --password="${DB_PASS}" --host="${DB_HOST}" "${DB_NAME}" < {}'
+
+  # Backwards compatibility for old way of importing.
+  elif [[ -f "${app_ref_data}"/db.sql.gz ]]; then
+    gunzip -c "${app_ref_data}"/db.sql.gz > "${tmp_ref_data}"-db.sql
+    pv -f "${tmp_ref_data}"-db.sql | drush sql-cli
```
ragnarkurmwunder, line 430: though the `pv` is good, and `pv -f` is forcing output even if not in the terminal... the output appears only when the command is completed. not sure if it is somehow possible to make it more realtime/dynamic.

charts/drupal/templates/_helpers.tpl
```diff
@@ -412,16 +410,28 @@ fi
...
+  fi
 
   # Clear caches before doing anything else.
-  if [[ $DRUPAL_CORE_VERSION -eq 7 ]] ; then drush cache-clear all;
+  if [[ "${DRUPAL_CORE_VERSION}" -eq 7 ]] ; then drush cache-clear all;
   else drush cache-rebuild; fi
```
ragnarkurmwunder, line 435: a bit confusing syntax.. not sure where the else belongs to.
suggesting to use extended version always, no shorthand, just for the sake of readability.
```
if [[ ... ]]; then
...
else
...
fi
```

ragnarkurmwunder (changes requested):
export consistency needs attention.

charts/drupal/templates/_helpers.tpl
Rade333: Hmm, to be on the safe side I renamed the directory to /tmp/reference-data-export/. There is no chance that such directory already exists :)

charts/drupal/templates/_helpers.tpl
Rade333: I wanted to make sure only *.sql files are being included in the archive, and ran into this problem when trying to use -C option: https://superuser.com/questions/266422/tar-c-with-a-wildcard-file-pattern

charts/drupal/templates/_helpers.tpl
Rade333: Well, this is the already existing code and just for backwards compatibility :)

charts/drupal/templates/_helpers.tpl
Rade333: I agree, I'll change to extended syntax. This was also existing code that I just did not update but will do it now.

Rade333:
I [removed the quotes](https://github.com/wunderio/drupal-project-k8s/pull/554/commits/385d7872e51135c36653893d7477ee9ee6eb30f7#diff-c4f491ae4ef2d83e98019bc3ac28e9ba383d1ec3573c7f25e65e1871f0f13821R384) wrapping the $nodata variables because they broke the mysqldump command when the variable was empty, see screenshots:
quotes
no-quotes

charts/drupal/templates/_helpers.tpl
ragnarkurmwunder: aha, i didnt know!

ragnarkurmwunder:
in this specific case i agree for not having quotes.

Rade333:
@ragnarkurmwunder the PR is now updated with a way of creating one file per database table. This should solve the locking-issue we talked about.

charts/drupal/templates/_helpers.tpl
```diff
@@ -373,23 +373,49 @@ done
...
-    IGNORED_TABLES="$IGNORED_TABLES $TABLE";
+  echo "Dump reference database."
+  dump_dir=/tmp/reference-data-export/
+  mkdir "${dump_dir}"
+
+  # Figure out which tables to skip the data for.
+  ignore_tables=''
+  for table in $(drush sql-query 'show tables;' | grep -E '{{ .Values.referenceData.ignoreTableContent }}'); do
+    ignore_tables="${ignore_tables} --ignore-table-data=${DB_NAME}.${table}";
   done
```
ragnarkurmwunder, line 384: Without loop
```shell
ignore_tables="$(
  drush sql-query 'show tables;' \
  | grep -E '{{ .Values.referenceData.ignoreTableContent }}' \
  | sed -e 's/^/--ignore-table-data=' \
  | tr '\n' ' '
)"
```

charts/drupal/templates/_helpers.tpl
```diff
@@ -373,23 +373,49 @@ done
...
 
   echo "Dump reference database."
-  mysqldump -u $DB_USER --password=$DB_PASS --host=$DB_HOST $IGNORE_TABLES $DB_NAME > /tmp/db.sql
-  mysqldump -u $DB_USER --password=$DB_PASS --host=$DB_HOST --no-data $DB_NAME $IGNORED_TABLES >> /tmp/db.sql
+  # The $ignore_tables variable cannot be quoted in the mysqldump command because if it's empty, the command will fail.
+  mysqldump --user="${DB_USER}" --password="${DB_PASS}" --host="${DB_HOST}" ${ignore_tables} "${DB_NAME}" > /tmp/db.sql
```
ragnarkurmwunder, line 388:
```shell
  { set +x; } 2>/dev/null
  mysqldump --user="${DB_USER}" --password="${DB_PASS}" --host="${DB_HOST}" ${ignore_tables} "${DB_NAME}" > /tmp/db.sql
  set -x
```

charts/drupal/templates/_helpers.tpl
```diff
@@ -373,23 +373,49 @@ done
...
+  last_table=$(find -type f -name 'table-*' | sort -n | tail -n1)
+  # Split last table file to extract mysqldump footer, which starts with a line including "@OLD_"
+  csplit --silent --prefix='last-' "${last_table}" '/@OLD_/'
+  # Replace $last_table with the version of it that has footer extracted from it
+  mv last-00 "${last_table}"
+  # Rename the extracted footer to "footer"
+  mv last-01 footer
+  # Prepend header and append footer to all table files, save them as .sql
+  for file in table-*; do
+    table_name=$(grep 'Table structure for table' ${file} | cut -d$'\x60' -f2)
```
ragnarkurmwunder, line 407: This works.

An alternative way:
```shell
table_name=$(grep 'Table structure for table' table-100 | sed -Ee 's/^.*`(.*)`.*$/\1/')
```

charts/drupal/templates/_helpers.tpl
```diff
@@ -373,23 +373,49 @@ done
...
+    cat header "${file}" footer > "${table_name}.sql"
```
ragnarkurmwunder, line 408: Nice quoting !!!

charts/drupal/templates/_helpers.tpl
```diff
@@ -412,17 +438,32 @@ fi
...
-  gunzip -c /app/reference-data/db.sql.gz > /tmp/reference-data-db.sql
-  pv -f /tmp/reference-data-db.sql | drush sql-cli
+  app_ref_data=/app/reference-data
+  tmp_ref_data=/tmp/reference-data
+
+  # New way of importing.
+  if [[ -f "${app_ref_data}/db.tar.gz" ]]; then
+    mkdir "${tmp_ref_data}"
+    tar -xzf "${app_ref_data}/db.tar.gz" -C "${tmp_ref_data}/"
+    find "${tmp_ref_data}/" -type f -name "*.sql" | xargs -P10 -I{} sh -c 'echo "Importing {}" && mysql -A --user="${DB_USER}" --password="${DB_PASS}" --host="${DB_HOST}" "${DB_NAME}" < {}'
```
ragnarkurmwunder, line 453: Quoting very good!

charts/drupal/templates/_helpers.tpl
```diff
@@ -373,23 +373,49 @@ done
...
+
+  previous_wd=$(pwd)
+  cd "${dump_dir}" || exit
+
+  # Split the dump to one file per table
+  csplit --silent --prefix='table-' /tmp/db.sql '/-- Table structure for table/'-1 {*}
+  # First file is the mysqldump header, rename it to "header"
+  mv table-00 header
+  # Find last table file
+  last_table=$(find -type f -name 'table-*' | sort -n | tail -n1)
```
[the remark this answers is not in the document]
ragnarkurmwunder: `sort -n` sorts by numbers, yes.
but not as humans normally would assume.

charts/drupal/templates/_helpers.tpl
Rade333: The -x was only in my debugs, it's not normally set.

charts/drupal/templates/_helpers.tpl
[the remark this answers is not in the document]
Rade333: Fixed by using 4 digits

PR #557, opened 2022-11-08
SLT-800: Improve Cron logging

ragnarkurmwunder:
* Generate statistics at the end of the execution
* Generate stats as JSON
* Capture job output, and export also in the JSON

The logs will look like this:
```JSON
{
  "cron_name":"drupal",
  "elapsed_time_s":0.96,
  "peak_mem_kb":222912,
  "exit_code":0,
  "output":"..."
}
```

The additional statistics can be turned on by adding `logging: format: json` entry into `silta.yml`:
```
logging:
  format: json
```


Rade333 (approved):
@ragnarkurmwunder approved, but please coordinate changes needed in Sumologic queries/dashboards/alerts, if any :)

Jancis (changes requested):
There needs to be a configuration parameter to enable this optionally. Logging format differs for other containers in deployment (mariadb, elasticsearch, other services that we don't control) and only few clusters have sumologic connected to them. This is a very specific vendor lock-in for some use cases. This can be merged, but only if it's not on by default.

Same applies to other PR's.

Rade333 (changes requested):

ragnarkurmwunder:
> and only few clusters have sumologic connected to them.

It is not a Sumologic-related change.
Sumo is just one of the tools that can easily take advantage of additional statistics.

> This is a specific implementation for some use cases.

Disagree with this. This could and actually should be the default option. Ideally we would like to perform Sumo queries like "show me the most mem-consuming cron tasks across the cluster" or "show me the most long-running cron tasks across the cluster".

Regardless, I will implement the suggestion.

I will also add one more parameter "total_mem_kb", so we can very easily find improper settings.

ragnarkurmwunder:
> Logging format differs for other containers ...

Not true. It only emits additional statistics. Look closely at `| tee ...`.

ragnarkurmwunder:
Cant find a simple way to emit `total_mem_kb`, skipping.

Rade333 (approved):

PR #558, opened 2022-11-08
SLT-802  Nginx Json Logs

ragnarkurmwunder:
See the description in [SLT-802](https://wunder.atlassian.net/browse/SLT-802).

Problem: Nginx logs are in a native format which is not very good for automated processing.
Solution: Convert the logs into JSON.

The JSON log can be turned on by the following settings:

```yml
logging:
  format: json
```

[SLT-802]: https://wunder.atlassian.net/browse/SLT-802?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Jancis (changes requested):
There is no description for this one, but I assume it's related to sumologic logging format. Same reason as https://github.com/wunderio/drupal-project-k8s/pull/557#pullrequestreview-1188032718, this can stay, but needs a feature gated and to be off by default

Rade333 (changes requested):

ragnarkurmwunder:
> There is no description for this one, ...

Added.

> but I assume it's related to sumologic logging format.

It's added not for Sumologic per se, but good automated log processing, regardless of tools used.

Added feature-gate by adding following settings into `silta.yml`:
```yml
nginx:
  json_logs: true
```

Jancis (approved):
If Ragnar was able to receive all log fields in desired format when format was set to json, this should do.

Rade333 (changes requested):

Rade333:
We agreed on a compromise where the original log format is kept, but a few additional details are added to the end of it.

@ragnarkurmwunder please create a new PR for that, I'm closing this one.

PR #559, opened 2022-11-08
SLT-803  Php-fpm Json Logs

ragnarkurmwunder:
See the description in [SLT-803](https://wunder.atlassian.net/browse/SLT-803).

> Problem: PHP-FPM logs are in a native format which is not very good for processing.
> Solution: Convert the logs into JSON.

The JSON log can be turned on by the following settings:
```yml
logging:
  format: json
```

[SLT-803]: https://wunder.atlassian.net/browse/SLT-803?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Jancis (changes requested):
There is no description for this one, but I assume it's related to sumologic logging format. Same reason as https://github.com/wunderio/drupal-project-k8s/pull/557#pullrequestreview-1188032718, this can stay, but needs to be feature gated and off by default

Rade333 (changes requested):

Jancis (approved):
Looks good! I believe global scope is not required in that context but it will work anyway.

Rade333 (changes requested):

Rade333:
We agreed on a compromise where the original log format is kept, but a few additional details are added to the end of it.

@ragnarkurmwunder please create a new PR for that, I'm closing this one.

PR #560, opened 2022-11-10
Change monolog config, Remove monolog restriction in CLI

Juhani-moilanen:
Include monolog in CLI 

The reason why it was not included was that monolog interfered with drush batch processing

There are 2 solutions
One is to use stderr instead of stdout https://github.com/drush-ops/drush/issues/4790
The other requires monolog version  2.2.0 where "drush" formatted is included https://www.drupal.org/i/3223751

Things to do/argue about
- to use the Drush formatted would need to add a check to settings.php for the monolog version
- Is there any reason to log in to stdout instead of stderr?
- Are there any arguments against the processors or minimum log level of "info"
- Change the filename to "silta.monolog.servises.yml" as there is no other config there

A simple drush command that uses batch is locale:check and makes this easier to test

### DIFF
--- a/charts/drupal/files/settings.silta.php
+++ b/charts/drupal/files/settings.silta.php
@@ -94,11 +94,5 @@ if (getenv('VARNISH_ADMIN_HOST')) {
 
 /**
  * Use our own services override.
- *
- * We don't include this when running on cli, for example when using drush,
- * because the output from monolog to stdout interferes with drush batch
- * processing, causing the process to die when drush tries to start a new batch.
  */
-if (PHP_SAPI !== 'cli') {
-  $settings['container_yamls'][] = 'sites/default/silta.services.yml';
-}
+$settings['container_yamls'][] = 'sites/default/silta.services.yml';
--- a/charts/drupal/files/silta.services.yml
+++ b/charts/drupal/files/silta.services.yml
@@ -1,9 +1,19 @@
 parameters:
   monolog.channel_handlers:
-     # Log to the stdout by default.
-     default: ['stdout']
+    # Log to the stdout by default.
+    default:
+      handlers:
+        - name: 'stream'
+          formatter: 'drush'
+  monolog.processors:
+    - 'message_placeholder'
+    - 'current_user'
+    - 'request_uri'
+    - 'ip'
+    - 'referer'
+    - 'filter_backtrace'
 
 services:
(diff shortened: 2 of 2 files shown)


Jancis:
This will need a round of tests (first release, release upgrades, cronjobs) and if it works, it can be merged. 

- Is there any reason to log into stdout instead of stderr?
  - If I recall it correctly, php containers don't differentiate stdout and stderr, they both are returned as stderr. It was created like this initially (https://github.com/wunderio/drupal-project-k8s/pull/197), but we can change it. We did change it for php-fpm (https://github.com/wunderio/drupal-project-k8s/pull/286) because php container does not send stdout anywhere (https://github.com/docker-library/php/issues/811), but never changed it for services.yml. We can use stderr for it :+1: . 

- Are there any arguments against the processors or minimum log level of "info"
  - Where is it defined? 

- Change the filename to silta.monolog.servises.yml as there is no other config there
  - I'll trust in Your drupal expertise and I think it's alright to change the filename to `silta.monolog.services.yml` if it avoids configuration conflicts.

Jancis:
This will need a round of tests (first release, release upgrades, cronjobs) and if it works, it can be merged. 

- Is there any reason to log into stdout instead of stderr?
  - If I recall it correctly, php containers don't differentiate stdout and stderr, they both are returned as stderr. It was created like this initially (https://github.com/wunderio/drupal-project-k8s/pull/197), but we can change it. We did change it for php-fpm (https://github.com/wunderio/drupal-project-k8s/pull/286) because php container does not send stdout anywhere (`https://github.com/docker-library/php/issues/811`), but never changed it for services.yml. We can use stderr for it :+1: . 

- Are there any arguments against the processors or minimum log level of "info"
  - Where is it defined? 

- Change the filename to silta.monolog.servises.yml as there is no other config there
  - I'll trust in Your drupal expertise and I think it's alright to change the filename to `silta.monolog.services.yml` if it avoids configuration conflicts.

Juhani-moilanen:
Had A discussion about the changes on 21.11.22
with @ragnarkurmwunder @hkirsman  and @tormi 

decided to split this change into smaller commits 
1. log to STDERR, Allow logging in CLI https://github.com/wunderio/drupal-project-k8s/pull/565
2. Add processors
3. Log level
4. File name

Will close this PR  in favor of the smaller changes

PR #562, opened 2022-11-11
Change units from bytes to bibytes

misterjoonas:
Not sure if the units should be changed somewhere else too.

### DIFF
--- a/charts/drupal/values.yaml
+++ b/charts/drupal/values.yaml
@@ -227,9 +227,9 @@ php:
   resources:
     requests:
       cpu: 5m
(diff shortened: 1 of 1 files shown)


charts/drupal/values.yaml
```diff
@@ -227,9 +227,9 @@ php:
   resources:
     requests:
       cpu: 5m
-      memory: 128M
+      memory: 122M
```
Jancis, line 230: I'm sure You meant to change these two as well :)

Jancis (approved):

PR #564, opened 2022-11-18
Ref data: Follow up to #554

Rade333:
A follow up to #554 

Changes proposed:
- remove duplicate echo line
- add echos when saving db dumps
- echo which db dump file is being imported


charts/drupal/templates/_helpers.tpl
```diff
@@ -388,7 +388,6 @@ if [[ "$(drush status --fields=bootstrap)" = *'Successful'* ]] ; then
     )"
   {{- end }}
 
-  echo "Dump reference database."
```
Rade333, line 391: This was a duplicate of line 376

Jancis (approved):

PR #565, opened 2022-11-21
Log to STDERR instead of STDOUT with monolog 

Juhani-moilanen:
### The Problem

Monolog is not included in CLI (cron commands) and we do not get cron logs in Sumo

### The Cause

Currently, this is due to `silta.services.yml` being excluded in `settings.silta.php`
```
/**
 * Use our own services override.
 *
 * We don't include this when running on CLI, for example when using Drush,
 * because the output from monolog to stdout interferes with the Drush batch
 * processing, causing the process to die when the Drush tries to start a new batch.
 */
if (PHP_SAPI !== 'cli') {
  $settings['container_yamls'][] = 'sites/default/silta.services.yml';
}
```

### The Goal

Get the Drupal cron messages in Sumo.

### The Solution

Use stderr instead of stdout

This was first discussed in this PR  https://github.com/wunderio/drupal-project-k8s/pull/560

Also added the Default log level of DEBUG so it is more visible
and changed the name of the default handler to "stream" ( as it uses the Streamhandler )
### Testing

A few commands that use batch processing 
`drush locale:check`
`drush en action` enabling any module

- Run commands in CLI that use drush batch processing 
- Check that there are no errors ( due to monolog spitting out logs)

STDOUT
image

STDERR
image

### DIFF
--- a/charts/drupal/files/settings.silta.php
+++ b/charts/drupal/files/settings.silta.php
@@ -95,10 +95,7 @@ if (getenv('VARNISH_ADMIN_HOST')) {
 /**
  * Use our own services override.
  *
- * We don't include this when running on cli, for example when using drush,
- * because the output from monolog to stdout interferes with drush batch
- * processing, causing the process to die when drush tries to start a new batch.
+ * Add monolog config
+ * We avoid monolog interfering with drush batch processing by logging to STDERR
  */
-if (PHP_SAPI !== 'cli') {
-  $settings['container_yamls'][] = 'sites/default/silta.services.yml';
-}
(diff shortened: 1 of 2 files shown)


Jancis (approved):
Thank You for this change! I don't know much about monolog module, but I suspect "stream" is only the name of custom stream wrapper and the actual output is still stderr which, again, turns out works since we moved to php 7.3+ (https://bugs.php.net/bug.php?id=71880)

operinko (approved):

Rade333:
Very good improvement @Juhani-moilanen! Thank you!

Rade333 (approved):
Thanks @Juhani-moilanen 👌

PR #610, opened 2023-04-13
Feature/cm newlines

Jancis:
This PR fixes newlines for configmap edits. Only snippet that is still broken, is `vcl_synth_500_vcl` and it's due to this special character: "🤔".

Adjusted include snippets (values.yaml overrides) are tested, they are compatible with custom configurations and return clean, expected result.

### DIFF
--- a/charts/drupal/templates/varnish-configmap-vcl.yaml
+++ b/charts/drupal/templates/varnish-configmap-vcl.yaml
@@ -15,9 +15,9 @@ data:
     backend prod1_http1 {
       .host = "{{ .Release.Name }}-drupal";
       .port = "80";
-      {{- .Values.varnish.backend_config | nindent 6 }}
+      {{ .Values.varnish.backend_config | nindent 6 | trim }}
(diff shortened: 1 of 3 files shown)


charts/drupal/templates/varnish-configmap-vcl.yaml
```diff
@@ -290,58 +290,58 @@ data:
...
+      }
       else {
         set req.http.X-Forwarded-For = client.ip;
         set req.http.X-Real-IP = client.ip;
       }
-      
+
       # Only deal with "normal" types
       if (req.method != "GET" &&
           req.method != "HEAD" &&
```
Rade333, line 347: Here there's still 4 space indentation

Rade333 (changes requested):
Found one place where there still was too much indentation

Jancis:
updated and tested, configuration still has correct newlines

Rade333 (approved):

PR #628, opened 2023-05-26
SLT-855: Allow DRUSH_OPTIONS_URI override

mikaelkundert:
These changes should allow projects to set `DRUSH_OPTIONS_URI` PHP environment variable without causing deployment error.

If `DRUSH_OPTIONS_URI` is not set, the output manifest should remain with the default/fallback value which uses `drupal.domain` value.

Note: I'm uncertain of how to test this locally, so I wish that someone is able to actually run these changes and confirm them working.


Jancis (approved):
Looks great, even has tests provided with it!

PR #629, opened 2023-06-05
Fastly Signal Sciences WAF support

Jancis:
Allows enabling Fastly Signal Sciences WAF agent for inline request reporting and blocking. This PR requires nginx image change (https://github.com/wunderio/silta-images/pull/101). 

Tests can be done by forking the branch and using it without last (cleanup) commit. Contains secret with Sigsci test keys and defines custom image (since nginx image changes ain't merged yet). 

Source: https://docs.fastly.com/signalsciences/install-guides/kubernetes/kubernetes-agent-module/


Rade333 (approved):
Tested as per PR instructions, works like a charm!

PR #630, opened 2023-06-06
SLT-808: Add value to allow override of DRUSH_OPTIONS_URI

Juhani-moilanen:
added `drushUri` value to `values.yml`
This allows developers to override the DRUSH_OPTIONS_URI in the template

## Testing
- SSH into current env and run `drush st` check that the `SITE URI` is correct
- create or check[ test branch](https://github.com/wunderio/drupal-project-k8s/tree/feature/SLT-808-test) SSH into it run `drush st` and check that the `SITE URI` is the same as set in `values.yml`

## Misc?

Should we actually check the `php.env.DRUSH_OPTIONS_URI` value for this?

### DIFF
--- a/charts/drupal/templates/_helpers.tpl
+++ b/charts/drupal/templates/_helpers.tpl
@@ -180,7 +180,7 @@ imagePullSecrets:
 - name: RELEASE_NAME
   value: "{{ .Release.Name }}"
 - name: DRUSH_OPTIONS_URI
-  value: "http://{{- template "drupal.domain" . }}"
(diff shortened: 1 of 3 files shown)


Jancis (changes requested):
Looks clean, thank You! Contextually, I'd suggest putting it under `php` section, right next to `hashsalt`, but looks like this is a duplicate of https://github.com/wunderio/drupal-project-k8s/pull/628/files and that PR provides basic / more common method of env variable override so I propose closing this one.

Rade333:
Closing in favor of #628

PR #632, opened 2023-06-07
SLT-775: Allow releaseMinAge to be annotated to ingress

Juhani-moilanen:
"Expose" releaseMinAge to developers to allow them to control when the deployment will be Downscaled
-- Still needs change in the[ downscaler repo](https://github.com/wunderio/silta-downscaler) to use the value 

## ToDo 
Test with downscaler


Jancis (approved):
Looks good, thank You!

PR #642, opened 2023-07-07
SLT-800:  Cron logs (JSON)

ragnarkurmwunder:
Based on the experience with earlier cron JSON here is a follow-up PR.

* Remove the feature gating, it is too much hassle for devs, and stops Wunder from reaping the benefits.
* Due to this, `logging: format' has now two options, `plain` and `JSON`, the latter being the default.
* Handle large cron outputs where a job goes rogue. Truncate the middle part.
* Append a newline at the end of the cron output if it is missing. This is needed because JSON would otherwise be in the same line.
* Add delimiters to be able properly to extract JSON from the log line which may contain other junk too.
* Add the size of the real output to JSON, otherwise, it cannot be guessed from log consumers.
* Add a detailed explanation for configuration parameters.


Rade333 (approved):

charts/drupal/values.yaml
```diff
@@ -791,5 +791,10 @@ signalsciences:
...
 logging:
-  # Currently supported formats are "default" and "json".
-  format: default
+  # Currently supported formats are:
+  # - "plain"
+  #   - output everything as it runs in real time.
+  # - "json" (default)
+  #   - output everything as it runs in real time.
+  #   - create additional overview-log-entry at the end including the cron job output.
+  format: json
```
Rade333, line 800: The changing of default value will still need a bit of discussion, let's do that on Slack.

Rade333 (changes requested):

charts/drupal/values.yaml
```diff
@@ -797,9 +797,3 @@ signalsciences:
     limits:
       cpu: 40m
       memory: 300Mi
-
-# Logging configurations for services that support them.
-# BETA feature, currently only supported by drupal-cron.
-logging:
-  # Currently supported formats are "default" and "json".
-  format: default
```
Rade333, line 805: As we discussed do not remove this key from values.yaml or the schema. Deployments that have it will fail if you do, see https://app.circleci.com/pipelines/github/wunderio/drupal-project-k8s/5448/workflows/22d0526d-0c52-4643-a105-f907b3535a47/jobs/17480

Let's do the removal in a separate PR with proper deprecation notices etc etc

Rade333 (changes requested):

Rade333 (commented):
Okay so I did the required changes myself and now the PR is ready BUT I just want to check @ragnarkurmwunder , you are aware that this change will increase the log size with around 30%, at least in this example?

I know you have the cap of 8192 and the output gets truncated, but still, won't this increase impact the logging costs quite much?

Feature branch:
feature-branch

Master branch:
master-branch

Comparison:
comparison

Rade333:
Closing due to inactivity.

PR #643, opened 2023-07-19
Feature/backup storageclass

Jancis:
creates backup and referenceData PersistentVolume only when silta-shared storageclass is used (since silta-shared does not provision volumes).


Rade333 (changes requested):
Nice addition! Please do the same for shell-volume as well.

Jancis:
shell volume conditional added

Rade333 (approved):

PR #644, opened 2023-07-20
SLT-768: Add nginx and php status pages

Jancis:
### ! Approve this first: https://github.com/wunderio/silta-images/pull/112

Couple status pages for nginx, enabled with `nginx.status_page.enabled` (`true` by default). Requires special image. Allows access to status pages only from `nginx.noauthips`.

**nginx status page (`/nginx_status`):**
Ekrānattēls 2023-07-20 13-17-55

**nginx virtualhost traffic status page (`/nginx_vts`):**
Ekrānattēls 2023-07-20 13-14-23

Also provides status page for php fpm (`/php_status`), enabled with `php.fpm.status_page.enabled`.
```
pool:                 www
process manager:      dynamic
start time:           10/Aug/2023:11:59:56 +0000
start since:          176
accepted conn:        24
listen queue:         0
max listen queue:     0
listen queue len:     511
idle processes:       1
active processes:     1
total processes:      2
max active processes: 1
max children reached: 0
slow requests:        0
```


charts/drupal/values.yaml
```diff
@@ -202,6 +202,9 @@ nginx:
   #   paths:
   #     - "^/sites/[^/]+/files/(?:css|js)/"
 
+  status:
```
Rade333, line 205: Let's change this key to `status_page` and add a brief comment above it to describe what this is.

Rade333 (changes requested):
Seems to work nicely! See my comment regarding the key though.

charts/drupal/values.yaml
Jancis: done. did the same for `php.fpm.status_page.enabled` to make it uniform.

Rade333 (approved):

PR #647, opened 2023-07-24
Feature/disable sigsci on probe

agnis-mateuss:
Remove robots.txt, _ping.php from paths scanned by SignalSciences.
Also removed redundant nginx config.


Rade333 (approved):
LGTM

PR #648, opened 2023-07-24
Change svc type

agnis-mateuss:
These services are not consumed out of the cluster, lets not expose them.


Rade333 (changes requested):
Thanks @agnis-mateuss, please change also the type for the shell service unless you see a reason not to.

Rade333 (approved):

PR #652, opened 2023-08-18
SLT-872 Cron resources logging

ragnarkurmwunder:
# Description
Enhance cron logging. Add additional fields to the JSON log:
* CPU time spent
* Memory request
* Memory limit
* CPU request
* CPU limit

# Reasons
* In SumoLogic it is possible to show users the resource limits and actual usage
* Automatic alerts about unreasonable resource specification

# References
* https://wunder.atlassian.net/browse/SLT-872
* https://github.com/wunderio/drupal-project-k8s/pull/557
* https://github.com/wunderio/drupal-project-k8s/pull/642

# Status of PR

This PR is branched off from this PR: https://github.com/wunderio/drupal-project-k8s/pull/642 which is not merged yet. So, therefore comparing this against `master` does not provide a clear picture of the changes in this PR.

There are minor things I'm not happy about, but not sure if we should go further with them or not.

* Instead of `{kernel_cpu_time_s: 0.43, user_cpu_time_s: 0.08}` we should have `{cpu_time_s: 0.51}`
* Instead of `{memory_request: "256M", memory_limit: "512M"}` we should have `{memory_request_mb: 256, memory_limit_mb: 512}`
* Instead of `{cpu_request: "300m", cpu_limit: "500m"}` we should have `{cpu_request_m: 300, cpu_limit_m: 500}`
* Undefined values should be `0` (integer, not string).

# Example
The image shows how many seconds of CPU were spent, and how much wall clock time it took. 
We can see that by allocating more CPU it could run many times faster, unless it is an io-intensive job that awaits mostly for the io resources. So the CPU in this particular case is under-allocated. The bars should be roughly the same.
Screenshot 2023-08-18 at 16 34 25

Here is an example of logs:
Screenshot 2023-08-18 at 16 46 29


Rade333:
It was decided that log is not the place to collect this data as there are other tools for that, closing PR.

PR #659, opened 2023-10-02
silta v1 tags

Jancis:
image version adjustments for Silta 1.0 initiative
Note: don't forget to sync `silta-release` library chart to charts repository!


Rade333 (changes requested):
As discussed, let's pin only major version for silta-release.

PR #662, opened 2023-10-30
SLT-901: Added cronjob defaults and updated cronjob resource inheritance logic

Jancis:
Allows setting cronjob defaults like resources or nodeselector. This PR also updates the inheritance logic used for assigning resource definitions for cronjobs, so that it works in the same way as it does in the frontend chart.

### DIFF
--- a/charts/drupal/templates/drupal-cron.yaml
+++ b/charts/drupal/templates/drupal-cron.yaml
@@ -98,23 +98,35 @@ spec:
                 exit "$exit_code"
 
             resources:
-              {{ if $job.resources }}
+              {{- if $job.resources }}
               {{- $job.resources | toYaml | nindent 14 }}
-              {{ else }}
+              {{- else }}
+              {{- if ($.Values.php.cronJobDefaults).resources }}
+              {{- $.Values.php.cronJobDefaults.resources | toYaml | nindent 14 }}
+              {{- else}}
               {{- $.Values.php.resources | toYaml | nindent 14 }}
+              {{- end }}
               {{- end }}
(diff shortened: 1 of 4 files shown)


charts/drupal/templates/drupal-cron.yaml
```diff
@@ -98,24 +98,36 @@ spec:
...
+              {{- if $job.resources }}
               {{- $job.resources | toYaml | nindent 14 }}
-              {{ else }}
+              {{- else }}
+              {{- if ($.Values.php.cronJobDefaults).resources }}
+              {{- $.Values.php.cronJobDefaults.resources | toYaml | nindent 14 }}
+              {{- else}}
               {{- $.Values.php.resources | toYaml | nindent 14 }}
               {{- end }}
+              {{- end }}
```
Rade333, line 109: Hmm, I was thinking that cronJobDefaults would be merged with job.resources. In a similar way how the logic for exposeDomainsDefaults works.

That would be more consistent with the way overriding chart values works overall.

Any objections or concerns if do it that way?

Rade333 (dismissed):
Tested and this works, but I started a discussion in a code-comment about the defaults-logic :)

charts/drupal/templates/drupal-cron.yaml
Rade333: I was now checking https://github.com/wunderio/frontend-project-k8s/pull/102 and noticed that it works exactly the way I meant. Let's change also this Drupal PR :)

Rade333 (approved):

PR #675, opened 2023-12-05
SLT-898: Redefine default cron schedules for different envs

Rade333:
Proposed changes:
- Adds a new silta-master.yml, which will be used in master env
- Defines new default cron schedules separately for feature envs, master env and production env. The point here is to have cron jobs run very seldomly in feature envs unless specifically needed to run more frequently.
- Refactor circleci config by using more anchors, adding more comments and removing references to obsolete branches

If accepted, these changes will also be made to the drupal-project template.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -164,7 +164,7 @@ workflows:
       - site-test: &site-test
           name: "Site test"
           requires:
-            - "build-deploy"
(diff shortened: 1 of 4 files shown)


Jancis (approved):
You might want to recreate this PR for drupal-project template

PR #682, opened 2024-01-10
Set vfs-cache-mode to "off" for backup volumes

Jancis:
This sets [vfs-cache-mode](https://rclone.org/commands/rclone_mount/#vfs-file-caching) to off for silta-shared backup pvc mounts and changes pvc & pv names to avoid immutable field issue with existing deployments.

Should solve backup unreliability issue where rclone process is stopped before it has fully copied file from .cache folder to remote.

Testing:
```yaml
backup:
  enabled: true
  schedule: '0 * * * *'
  retention: 2
```

### DIFF
--- a/charts/drupal/templates/backup-volume.yaml
+++ b/charts/drupal/templates/backup-volume.yaml
@@ -3,9 +3,9 @@
 apiVersion: v1
 kind: PersistentVolume
 metadata:
-  name: {{ .Release.Name }}-{{ $.Release.Namespace | sha256sum | trunc 7 }}-backups
+  name: {{ .Release.Name }}-{{ $.Release.Namespace | sha256sum | trunc 7 }}-backup
   labels:
-    name: {{ .Release.Name }}-{{ $.Release.Namespace | sha256sum | trunc 7 }}-backups
+    name: {{ .Release.Name }}-{{ $.Release.Namespace | sha256sum | trunc 7 }}-backup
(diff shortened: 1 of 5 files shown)


charts/drupal/templates/backup-volume.yaml
```diff
@@ -16,16 +16,17 @@ spec:
   {{- if .Values.backup.csiDriverName }}
   csi:
     driver: {{ .Values.backup.csiDriverName }}
-    volumeHandle: {{ .Release.Namespace }}-{{ $.Release.Namespace | sha256sum | trunc 7 }}-backups
+    volumeHandle: {{ .Release.Namespace }}-{{ $.Release.Namespace | sha256sum | trunc 7 }}-backup
     volumeAttributes:
       remotePathSuffix: /{{ .Release.Namespace }}/{{ .Values.environmentName }}/backups
+      vfs-cache-mode: "off"
```
Rade333, line 22: So does this new line override [this one](https://github.com/wunderio/csi-rclone/blob/master/pkg/rclone/nodeserver.go#L247), or what is the current default?

Jancis: yes, it excludes flags when override is defined

https://github.com/wunderio/csi-rclone/blob/master/pkg/rclone/nodeserver.go#L294-L305

Rade333 (approved):
Looking good

PR #687, opened 2024-01-16
removes port from local redirects

Jancis:
Removes port from local redirect. Otherwise, users would get redirected to :8080 when using redirects without hostnames.

http://nginx.org/en/docs/http/ngx_http_core_module.html#port_in_redirect
http://nginx.org/en/docs/http/ngx_http_core_module.html#server_name_in_redirect


Rade333 (approved):

Rade333:
As we found that the remove port directive already exists in this chart, this PR is no longer needed, closing.

PR #689, opened 2024-01-18
Allow unsetting cronjobs

Jancis:
Allows unsetting default cronjobs or cronjobs from shared cronjob list for release-suffixed deployments


Rade333 (approved):
This is a good addition!

PR #695, opened 2024-02-13
SLT-810: Overwrite PHP settings for individual cronjobs

agnis-mateuss:
Adds an option to specify additional PHP ini settings for cronjobs.
Example:
```php:
  cron:
    drupal:
      php_ini: |
        [Date]
        date.timezone = Europe/Riga
```
php_ini key creates a new entry in drupal-configmap and mounts it as a file in the specified cronjob pod.

This PR is dependent on Helm [3.14](https://github.com/wunderio/silta-images/pull/173) images being released beforehand.

### DIFF
--- a/charts/drupal/Chart.yaml
+++ b/charts/drupal/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: drupal
-version: 1.4.1
+version: 1.4.0
 dependencies:
 - name: mariadb
   version: 7.5.x
--- a/charts/drupal/Chart.yaml
+++ b/charts/drupal/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: drupal
-version: 1.4.0
+version: 1.4.1
 dependencies:
 - name: mariadb
   version: 7.5.x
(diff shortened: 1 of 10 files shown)


charts/drupal/values.schema.json
```diff
@@ -251,6 +251,7 @@
               "command": { "type": "string" },
               "backoffLimit": { "type": "integer" },
               "nodeSelector": { "type": ["object", "null"] },
+              "php_ini": { "type": "string" },
```
Rade333, line 254: Please add this new key also to the default drupal cron job in values.yaml, even if it would be just an empty value, but so that people might notice that it is available. A brief comment would also be nice.

charts/drupal/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: drupal
-version: 1.4.0
+version: 1.4.1
```
Rade333, line 3: This is okay, but for future PRs no need to bump chart version as that is part of the release process :)

Rade333 (changes requested):
Looks good, left minor comments. Also could you please check why this validation fails in my test branch? https://app.circleci.com/pipelines/github/wunderio/drupal-project-k8s/6019/workflows/007a6d91-2cac-4b91-b023-e45d68300309

PR #702, opened 2024-03-11
Sync nginx client_header_timeout and client_body_timeout with php max_execution_time

artisba:
In case there is a need to increase PHP's `max_execution_time` value the current implementation doesn't have any effect unless `client_body_timeout` / `client_header_timeout` in nginx config are also adjusted accordingly.

This allows both to override these values and keep them in sync with `max_execution_time` while keeping the current default of `60` seconds.

### DIFF
--- a/charts/drupal/templates/drupal-configmap.yaml
+++ b/charts/drupal/templates/drupal-configmap.yaml
@@ -40,6 +40,7 @@ data:
     request_order = GP
     post_max_size = {{ .Values.php.php_ini.post_max_size }}
     upload_max_filesize = {{ .Values.php.php_ini.upload_max_filesize }}
+    max_execution_time = {{ .Values.php.php_ini.max_execution_time }}
 
     [Date]
     date.timezone = Europe/Helsinki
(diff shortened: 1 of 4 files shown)


charts/drupal/templates/drupal-configmap.yaml
```diff
@@ -40,6 +40,7 @@ data:
     request_order = GP
     post_max_size = {{ .Values.php.php_ini.post_max_size }}
     upload_max_filesize = {{ .Values.php.php_ini.upload_max_filesize }}
+    max_execution_time = {{ .Values.php.php_ini.max_execution_time }}
```
Rade333, line 43: max_execution_time is already set on line 34. Please replace that line with this new one.

Rade333 (changes requested):

charts/drupal/templates/drupal-configmap.yaml
Rade333: Fixed this myself

Rade333 (approved):

PR #715, opened 2024-08-15
Disable privileged init container

Jancis:
## Blocked by https://github.com/wunderio/charts/pull/444

Disables sysctl init container, removes privileged pod definition from template.
This change disables `vm.max_map_count` sysctl modification.
```
command: ["sysctl", "-w", "vm.max_map_count={{ .Values.sysctlVmMaxMapCount}}"]
```

### DIFF
--- a/charts/drupal/values.yaml
+++ b/charts/drupal/values.yaml
@@ -658,9 +658,6 @@ elasticsearch:
       memory: 1Gi
   sysctlInitContainer:
     enabled: false
-  esConfig:
-    elasticsearch.yml: |
-      node.store.allow_mmap: false
   nodeAffinity:
     preferredDuringSchedulingIgnoredDuringExecution:
     - weight: 100
--- a/charts/drupal/values.yaml
+++ b/charts/drupal/values.yaml
@@ -632,9 +632,7 @@ elasticsearch:
   minimumMasterNodes: 1
   maxUnavailable: 0
   clusterHealthCheckParams: 'wait_for_status=yellow&timeout=1s'
-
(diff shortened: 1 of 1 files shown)


Rade333:
Approved but cannot be merged before https://github.com/wunderio/charts/pull/444 since we need the new Daemonset from there

Jancis:
If the `silta-cluster` can't be upgraded yet, here's a temporary daemonset that will take care of it until release. Identical name is intentional, so that `silta-cluster` chart upgrade would fail and require removal of this temporary fix.

Apply with `kubectl apply -f`
```yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: silta-cluster-ds
  namespace: silta-cluster
spec:
  selector:
    matchLabels:
      name: silta-cluster-ds
  template:
    metadata:
      labels:
        name: silta-cluster-ds
    spec:
      containers:
      - command:
        - sleep
        - infinity
        image: docker.io/bash:5
        imagePullPolicy: IfNotPresent
        name: sleep
      initContainers:
      - command:
        - /usr/local/bin/bash
        - -e
        - -c
        - echo 262144 > /proc/sys/vm/max_map_count
        image: docker.io/bash:5
        imagePullPolicy: IfNotPresent
        name: max-map-count-setter
        resources:
          limits:
            cpu: 100m
            memory: 32Mi
        securityContext:
          privileged: true
          runAsUser: 0
      restartPolicy: Always
      securityContext: {}
```

Jancis:
let's not roll this one out yet and disable this per-cluster until we know all clusters are actually upgraded.

PR #724, opened 2024-11-12
Add cronjob name as label

Rade333:
Changes proposed:
- Add the name (key) of the cronjob as a label. This is helpful when dealing with multiple cronjobs that have shortened/hashed names

Monosnap k9s 2024-11-12 12-00-09


Jancis (approved):
Good idea!

PR #726, opened 2024-11-22
Add nodeSelector support for backup jobs

Rade333:
Changes proposed:
- Add nodeSelector support for backup jobs
- Set the default value to an empty object

Here are screenshots from test:

**Backups enabled, using default value for nodeSelector:**

Monosnap k9s 2024-11-22 11-20-06
Monosnap k9s 2024-11-22 11-20-33

**Backups enabled, using overridden value for nodeSelector:**

Monosnap k9s 2024-11-22 11-19-28
Monosnap k9s 2024-11-22 11-16-18


evilfurryone (approved):
looks good. We can then either add the backup nodeselector value into the values file or better, override it in circleci via the context override.

PR #733, opened 2025-01-21
MariaDB 11 server

operinko:
…-fpm image as well, but we'll see.


Rade333 (changes requested):
Let's change this to 10.11 instead, as that is what our Silta images are using

PR #736, opened 2025-02-18
SLT-1136: Drupal chart: Use JSON as the default formatter for application logs when handled by monolog

k4lv15:
Adjust the default monolog channel handler:

- set formater to JSON to have structured Drupal logs
- specify default log processors
- override monolog's filter_backtrace processor to remove redundant log data which makes sense in Drupal UI but not on stderr


Jancis:
Works as expected, can see structured log picked up by otel and displayed in logging system.
Not excited that the log format is json by default, but worth it if it makes it easier for developers to navigate logs.

Rade333 (approved):

PR #741, opened 2025-04-01
SLT-1103: Drupal chart: Add support for mailpit, deprecate mailhog

k4lv15:
**Motivation:**
[Mailhog](https://github.com/mailhog/MailHog) is an email service we’re using in our development environments (local & Silta non-production) to test out if email functionalities are working properly. Mailhog [is not maintained anymore](https://github.com/mailhog/MailHog/issues/442#issuecomment-1493415258) and is causing issues for developers. Therefore, we need to replace it with some alternative. Our current pick is [Mailpit](https://mailpit.axllent.org/).

Changes proposed:
- Add mailpit dependency, enable it, add default value overrides
- Update chart templates with needed changes for mailpit to work
- Update unit tests to run against mailpit instead of mailhog
- Keep mailhog in chart but add deprecation notices

How to test:
1. Install the updated chart and see that mailpit works (`helm install drupal-release charts/drupal --set mailpit.enabled=true` or test here: https://app.circleci.com/pipelines/github/wunderio/drupal-project-k8s/6873/workflows/20d2da53-257e-4bc2-8d2e-85bbb7c3eae2/jobs/22713). Test the following:
    - validation goes through (i.e., unit tests pass: `helm unittest ./charts/drupal`)
    - mailpit URL is listed in the release notes and it works
    - the legacy `/mailhog` path redirects to `/mailpit`
    - emails sent from Drupal (i.e., after password reset or new account creation) end up in mailpit, mailpit's UI works, i.e, you can mark emails as read, delete them, download etc.
2. See that updated installation checks work, i.e., try to install chart with both `mailpit` and `mailhog` enabled, see that the installation fails (or see screenshot below)
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -142,6 +142,7 @@ workflows:
                 name: Add helm repositories and build local chart
                 command: |
                   helm repo add elastic https://helm.elastic.co
+                  helm repo add jouve https://jouve.github.io/charts/
                   helm repo add codecentric https://codecentric.github.io/helm-charts
                   helm repo add percona https://percona.github.io/percona-helm-charts/
                   helm dependency build ./charts/drupal
--- a/charts/drupal/Chart.lock
+++ b/charts/drupal/Chart.lock
@@ -16,12 +16,12 @@ dependencies:
   version: 5.1.0
 - name: mailpit
   repository: https://jouve.github.io/charts/
-  version: 0.23.1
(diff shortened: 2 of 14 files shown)


tormi (approved):
RTBC. Nice work, thanks @k4lv15 ! 👍 

Screenshot 2025-04-01 at 22 13 07

PR #770, opened 2025-09-15
SLT-1232: Add storage.silta/storage-path annotation to Solr's volumeClaimTemplates

k4lv15:
tba


Rade333 (approved):
Nice addition, thanks @k4lv15 !

